PR #1, opened 2022-08-08
Use taccsite_custom dir for static files.

jarosenb:
When the image is built, static files live under `/code/taccsite_custom`.


wesleyboar:
I tested this successfully, locally.

_I did not use `make` commands to test, though. Those needed fixing. See #2._

wesleyboar (approved):

PR #28, opened 2022-10-19
Task/fp 1838 apcd form updates → FP-1888 scripts to save or discard

wesleyboar:
## Important

Now that we don't use plugin, some scripts are not used, but could prove useful for a form that does use the plugin.

## Overview

Save snippets that are ~~still used~~ [possibly useful] for future plugin-powered form.

## Related

- [FP-1838](https://jira.tacc.utexas.edu/browse/FP-1838)
- https://github.com/TACC/Core-CMS-Custom/pull/35

## Changes / Testing / UI

Abandoned.


wesleyboar:
Removal of scripts (and styles) is in #35. I can refer to it if I need to bring back any more scripts. The ones here though were the most likely candidates to retain and reuse.

PR #34, opened 2022-10-26
Task/FP-1908 - Set default value for registration_status

Strmiska:
## Overview

Minor changes to the Registration Form

## Related

- [FP-1908](https://jira.tacc.utexas.edu/browse/FP-1908)

## Changes

- Set a default value for Registration Status to "Received"
- Prevent form caching on reload/back navigation (prevent the same form data from being reused)
- Include a missing parenthesis for Entity Being Registered

## Testing

1. Fill out some fields on the [form](http://localhost:8000/register/request-to-submit/) (no need to submit)
2. Navigate to another page, and then use the back button to try returning to the form
3. The form should be cleared
4. No means to test database submissions locally at the moment


edmondsgarrett (commented):
Caching appears to still be working.

Strmiska:
> Caching appears to still be working.

Ah yeah no worries, I've looked this up and it seems to be a pretty complicated issue to tackle with just Django. Seems like the issue still depends on the browser type even when creating custom decorators. It's not the primary goal of this PR anyways, so we may have to tackle it in another task.

apcd-cms/src/apps/registrations/views.py
```diff
@@ -16,6 +17,7 @@
 
 
 class SubmissionFormView(View):
+    @never_cache
```
wesleyboar, line 20: If cache is not working (I trust @edmondsgarrett), and [not the main goal of the PR](https://github.com/TACC/Core-CMS-Custom/pull/34#issuecomment-1309482551), then should it be removed? (So as not to introduce a non-working feature that may become a misdirection[^1] or a bug[^2].)

[^1]: Someone could think this solves cache (but it doesn't) and respond accordingly (but incorrectly).
[^2]: This may result in non-intuitively inconsistent cache behavior.

wesleyboar (commented):
Hrm. I'm keen on everything except the cache issue, which seems unresolved. I've left thoughts in comment on `@never_cache`.

apcd-cms/src/apps/registrations/views.py
Strmiska: That's a good point. I'll pull it out for now and make note of it in the JIRA ticket.

fnets (approved):
LGTM. Haven't had a chance to test locally, but this is pretty simple

PR #38, opened 2022-10-29
Create demdata Github Action

fnets:
## Overview
This will create the workflow file for democratizing data and push up the containers for CMS commits. Basically a clone from the APCD one.

## Related

- https://github.com/TACC/Core-CMS-Custom/pull/39

## Changes

Init Commit.

## Testing

1. Run the workflow.

## UI

N/A

### DIFF
--- a/.github/workflows/demdata-cms.yml
+++ b/.github/workflows/demdata-cms.yml
@@ -1,0 +1,41 @@
+name: Democratizing Data CMS Builds 
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+    paths:
+      - 'demdata-cms/**'
+
+jobs:
+  build_commit:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: demdata-cms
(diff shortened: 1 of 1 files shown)


wesleyboar (approved):
Because this is for a directory that does not exist until #39 is merged, I suggest **either** delaying it until #39 is merged **or** merging this into `demdatainit` (you can just the the base of this PR to that branch, and merge).

wesleyboar:
Notice: I tweaked the description for accuracy, mostly.

But "Testing"... I do not know how to run this new workflow successfully without the directory in place. If you have a way, please teach 🙂.

_P.S. [It seems the workflow can be run on a branch before `main` though, by pushing a commit](https://stackoverflow.com/a/63568873/11817077); this is new to me._

fnets:
> But "Testing"... I do not know how to run this new workflow successfully without the directory in place. If you have a way, please teach 🙂.

I don't think you actually can, as far as google tells me. You're right, we need to merge this before we can test.

PR #43, opened 2022-11-08
Task/FP-1837 - Set APCD user roles on Django user model

Strmiska:
## Overview

The changes in this PR will allow the collected user roles from the APCD database to be store in the Django User model on the portal. In order for these changes to work, the user will need access to the database and we will have to update the `LOGIN_REDIRECT_URL` to `/apcd-login/get-roles/`.

## Related

- [FP-1837](https://jira.tacc.utexas.edu/browse/FP-1837)

## Changes

- Create a new endpoint for remote login so we can gather APCD user roles.
- Create APCD roles in the CMS as groups 
- Update portal settings for `LOGIN_REDIRECT_URL`

## Testing

Currently deployed in PPRD
1. Open incognito session
2. Log into the portal
3. Log into the CMS admin panel
4. Find your user and check that you have 1 APCD group assigned to you. You will have an APCD group assigned to you if you have a role set within the APCD database


wesleyboar (commented):
Skimmed. Looks good. Feel free to deploy to QA server to test if no one else has anything up there. I do not. Testing this locally would be difficult (until one day, when the clouds have gone away and the flowers gleam).

fnets (approved):
LGTM! Good job

PR #53, opened 2022-11-28
fix(apcd-cms): improve reg. list table desc

wesleyboar:
## Overview

New description for registration table.

## Related

- [discussion with Design](https://tacc-team.slack.com/archives/CQUA4D5KJ/p1669667188616439)

## Changes

- changed text of template

## Testing & UI

I cheated—on just this one—by editing live markup.

description update

I would have just done this via the CMS, if I could.[^1]

[^1]: Ideally, this text is in the CMS, not the template. I'll fix, or create a ticket to fix this, one day.


edmondsgarrett (approved):
I like the brevity! Approved

PR #45, opened 2022-11-10
test(apcd-cms): fp-1837 limit menu visibility by group

wesleyboar:
## Overview

Implement custom logic in CMS nav template.

## Related

- [FP-1837](https://jira.tacc.utexas.edu/browse/FP-1837)

## Changes

- clone cms nav template, but use template tag
- add template tag with menu view permission logic

### Important

You must re-build and re-up containers after checking out this branch.

```
docker-compose -f docker-compose.dev.yml down && docker-compose -f docker-compose.dev.yml build --no-cache && docker-compose -f docker-compose.dev.yml up --force-recreate --detach
```

### Setup

As admin,
1. Create user groups:
    - `SUBMITTER_USER`
    - `SUBMITTER_ADMIN`
    - `APCD_ADMIN`
2. Create a test user that:
    - is "Active"
    - has "Staff status"
    - not "Superuser"
3. Create three pages that are visible in menu:
    - Home
    - Submissions (Advanced Settings: ID: `submissions`)
    - Administration (Advanced Settings: ID: `administration`)

### Steps

1. While not logged in, login as test user.
    <sup>In a private window, so you can check admin steps in your main window.</sup>
2. As admin, for test user, choose group `SUBMITTER_USER`.
3. As test user, load any CMS page.
4. Verify user sees this navigation:
    - Home
    - Submissions
5. For test user, choose group `SUBMITTER_ADMIN`.
6. Verify user sees this navigation:
    - Home
    - Submissions
7. For test user, choose group `APCD_ADMIN`.
8. Verify user sees this navigation:
    - Home
    - Submissions
    - Administration
9. For test user, choose **no** groups (i.e. un-assign `APCD_ADMIN` group).
10. Verify user sees this navigation:
    - Home

## UI

| Step 4 | Step 6 | Step 8 | Step 10 |
| - | - | - | - |
| Step 3 | Step 6 | Step 9 | Step 12 |

## Notes

**Next**: Add the same template logic in CMS, and add a template tag of the same name, but whose `simple_tag` always returns `True`, and that has documentation that the template tag may be overwritten (like in APCD). Then confirm that APCD is unaffected when using the new Core CMS image that has that.

PR #55, opened 2022-12-06
fix(templatetags): limit menu visibility + missing page id key = crash

wesleyboar:
## Overview

**Fix** that **if** any menu item does not have a `reverse_id`*†, **then** `limit_visibility_in_menu.py` will **crash**.

<sub>If a CMS page has no ID* set, it will still have a `reverse_id`, it's just blank. This fixes edge case of menu item **not having the attr/key**.†<sub>

<sub>* I.e. an "ID" in Advanced Settings.\
† E.g. CMS News category or article.</sub>

## Related

- copied from https://github.com/TACC/Core-CMS/pull/571
- adds safety to https://github.com/TACC/Core-CMS-Custom/pull/45

## Changes

- check for `reverse_id` before using it

## Testing

0. Load this branch on existing APCD setup.
1. Open any page of the site (not the CMS Administration).
2. Verify no crash (from newly introduced code).
3. Verify any of the test cases from [#45](https://github.com/TACC/Core-CMS-Custom/pull/45) still work.

## UI

N/A


wesleyboar:
Merged based on its unobtrusive and safety, and it having been tested via Core in #571.

PR #65, opened 2023-01-18
feat(apcd-cms): replace options on status filter change

wesleyboar:
⚠️ This is sample code. It has not been tested. It only has sample `<option>` elements. ⚠️

## Overview

Sample code to swap registration list options when status is changed.

## Related

…

## Changes

- added template options
- added function to swap options
- added call to new function

## Testing

0. Open registration list table.
1. Change filter value.
2. Verify actions change.
3. Verify table is **still** filtered.

## UI

…

### DIFF
--- a/apcd-cms/src/apps/admin_regis_table/templates/list_registrations.html
+++ b/apcd-cms/src/apps/admin_regis_table/templates/list_registrations.html
@@ -17,7 +17,7 @@
   <div class="filter-container">
     <div class="filter-content">
       <span><b>Filter by Status: </b></span>
-      <select id="statusFilter" class="status-filter" onchange="filterTableByStatus()">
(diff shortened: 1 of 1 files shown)


wesleyboar:
I misunderstood the goal. The code is useful reference in general, but is specific to the wrong template.

wesleyboar:
Also, @sophia-massie lead us toward a more compact way to do this:
1. List all possible options in the second dropdown.
2. On each option, identify the option of the first dropdown to which it belongs.
3. On select of an option in the first dropdown, loop through options in second drop-down.
4. For any (second dropdown) option that is **not** for the (first dropdown) option value, hide it.
5. For any (second dropdown) option that **is** for the (first dropdown) option value, un-hide it:
    - In case it was previously hidden.
7. Also, un-select whichever option was selected.

PR #72, opened 2023-01-23
Fp 1815 exception page 1 23

sophia-massie:
## Overview

…

## Related

…

## Changes

…

## Testing

1.

## UI

…


apcd-cms/src/apps/submissions/urls.py
```diff
@@ -1,8 +1,12 @@
...
 
 app_name = 'submissions'
 urlpatterns = [
     path('list-submissions/', SubmissionsTable.as_view()),
-]
\ No newline at end of file
+    path('exception/', ExceptionFormView.as_view(), name="exception"),
+    path('threshold-exception/', ExceptionThresholdFormView.as_view(), name="threshold-exception"),
+    path('other-exception/', ExceptionOtherFormView.as_view(), name="other-exception")
+]
```
shayanaijaz, line 12: I don't think the exception paths are needed here since you already included them through `exception/urls.py` file

sophia-massie: Ah good catch. So I tried to remove this from the submissions url but the the pages would not longer appear on that path. So instead, I removed the URL patterns from the exception urls.py since that is not longer needed. exceptions/urls.py did need at least one url pattern, so I left the main landing page /exception/ with name='index' and that seemed to work.

apcd-cms/src/apps/exception/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from apps.exception.views import ExceptionFormView
+from apps.exception.views import ExceptionOtherFormView
+from apps.exception.views import ExceptionThresholdFormView
+
+app_name = 'exception'
+urlpatterns = [
+    path('exception-request/', ExceptionFormView.as_view(), name="index"),
```
shayanaijaz, line 8: Can you try these changes to see if it would work having the exception urls in this file?

```suggestion
    path('exception-request/', ExceptionFormView.as_view(), name="index"),
    path('threshold-exception/', ExceptionThresholdFormView.as_view(), name="threshold-exception"),
    path('other-exception/', ExceptionOtherFormView.as_view(), name="other-exception")
```

apcd-cms/src/apps/submissions/urls.py
```diff
@@ -1,8 +1,11 @@
...
+from apps.exception.views import ExceptionFormView, ExceptionThresholdFormView, ExceptionOtherFormView
 
 app_name = 'submissions'
 urlpatterns = [
     path('list-submissions/', SubmissionsTable.as_view()),
-]
\ No newline at end of file
+    path('exception/', ExceptionFormView.as_view(), name="exception"),
+    path('threshold-exception/', ExceptionThresholdFormView.as_view(), name="threshold-exception"),
+    path('other-exception/', ExceptionOtherFormView.as_view(), name="other-exception")
```
shayanaijaz, line 10:
```suggestion

```

apcd-cms/src/apps/submissions/urls.py
```diff
@@ -1,8 +1,11 @@
 from django.urls import path
 from apps.submissions.views import SubmissionsTable
-from apps.admin_regis_table.urls import RegistrationsTable
+from apps.exception.views import ExceptionFormView, ExceptionThresholdFormView, ExceptionOtherFormView
```
shayanaijaz, line 3:
```suggestion
```

apcd-cms/src/apps/exception/urls.py
sophia-massie: This works for me locally not sure what the issue was on my end originally. Only thing is the path on line 8 needs to be `    path('exception/', ExceptionFormView.as_view(), name="index"),` instead. I'll push these changes up now. Thank you!

sophia-massie (commented):
Merged these changes. Thank you!

shayanaijaz (approved):
LGTM!

edmondsgarrett (approved):
LGTM!

PR #79, opened 2023-01-31
hotfix: Create submitter and edit registration not writing to database

edmondsgarrett:
## Overview
Fixes for edit registration and create submitter actions on list registration requests admin page
…

## Related

…

## Changes
- Added a missing value field for create submitter SQL operation command
- Added missing `cur` definition in update registration util function
- Reformatted current time call for update registration operation
…

## Testing

1. Will have to test on QA server, I've tested the revised util functions separately and they work.
2. Therefore, nothing to test locally.

## UI

…


sophia-massie (approved):
LGTM!

PR #90, opened 2023-02-06
View users

davidmtacc:
## Overview
Added "All" to dropdown list. Made Dropdown list display sorted alphabetically and not show duplicates.
Cleaned up by removing console logs and duplicate code in views.
…

## Related

…

## Changes

Added javascript to sort dropdown list and table.  Changed dropdown list to include ALL

## Testing

1. Tested using made-up data on local machine
2. 
Screenshot 2023-02-06 at 11 04 42 AM

## UI

…


apcd-cms/src/apps/view_users/views.py
```diff
@@ -25,61 +25,6 @@ def get_context_data(self, *args, **kwargs):
```
sophia-massie, line 25: Remove import datetime from get_context_data

sophia-massie, line 25: Pass user_content in from main view by passing it in as an argument to the get_context_data function like this:

def get_context_data(self, user_content = user_content, *args, **kwargs):

Then remove line 23

sophia-massie (changes requested):
Is there a way to change it so the role_id lists whether someone is a Submitter User, Submitter Admin, or APCD Admin instead of a number?

fnets (changes requested):
Looks like the `utils` function `get_usersIO` is deleted, but still used in the `view_users/views.py`

davidmtacc:
Updated to make changes requested.

apcd-cms/src/apps/view_users/views.py
```diff
@@ -16,81 +16,24 @@ def dispatch(self, request, *args, **kwargs):
            return HttpResponseRedirect('/')
         return super(ViewUsersTable, self).dispatch(request, *args, **kwargs)
 
-    def get_context_data(self, *args, **kwargs):
+    def get_context_data(self, user_content, *args, **kwargs):
```
sophia-massie, line 19:
```suggestion
    def get_context_data(self, user_content=user_content, *args, **kwargs):
```

sophia-massie (changes requested):
Some issues with responsive formatting when making the screen smaller that might affect cell phones. Table column headers become row headers which is fine but then some of the row headers disappear. Maybe responsive design is something to look over on a second pass once all the pieces are working. Everything else looks good after the change is made.

sophia-massie (approved):
May need to re-think how the table formats dates and role ID on a second pass. Also columns on right are squished compared to the email column. Everything is working as intended though and we can get with Tracy and UT Health for design direction.

edmondsgarrett (approved):
LGTM! Agreed with Sophia's notes on UI

apcd-cms/src/apps/view_users/views.py
```diff
@@ -101,7 +44,7 @@ def _set_modal_content(usr):
         
         
 
-        context['header'] = ['User ID', 'Email', 'Name', 'Organization', 'Role ID', 'Created', 'Updated', 'Notes']
+        context['header'] = ['User ID', 'Email', 'Name', 'Organization', 'Role', 'Active', 'Created', 'Updated', 'Notes']
         context['rows'] = []
         for user in user_content:
             user_view = [usr for usr in user_content if usr[1] == user[0]]
```
fnets, line 50:
```suggestion
```

apcd-cms/src/apps/view_users/views.py
```diff
@@ -16,81 +16,24 @@ def dispatch(self, request, *args, **kwargs):
...
                     'created_at': usr[5],
                     'updated_at': usr[6],
                     'notes': usr[7],
-                    
+                    'active': usr[8],
+                    'role_name': usr[9],
+    
                 }
 
         def _set_modal_content(usr):
```
fnets, line 39:
```suggestion
```

apcd-cms/src/apps/view_users/views.py
```diff
@@ -16,81 +16,24 @@ def dispatch(self, request, *args, **kwargs):
            return HttpResponseRedirect('/')
         return super(ViewUsersTable, self).dispatch(request, *args, **kwargs)
 
-    def get_context_data(self, *args, **kwargs):
+    def get_context_data(self, user_content=user_content, *args, **kwargs):
         context = super(ViewUsersTable, self).get_context_data(*args, **kwargs)
         actions = 'View'
```
fnets, line 21:
```suggestion
```

fnets (changes requested):
As far as I can tell, the parts I suggested to delete are no longer used.

fnets (approved):
LGTM!

PR #92, opened 2023-02-06
hotfix/create-extension-bug

shayanaijaz:
## Overview

Fix for data not saving to database after creating extension. Also tweaked queries that were pulling in the org names for a user. The org names displayed should be correct now.

## Related

…

## Changes

- Removed insertion of primary key when creating an extension (key auto increments)
- Tweaked queries to get org name from `apcd_orgs` table instead which should be a more accurate table to use to get organizations

## Testing

- No testing to be done locally. After this gets deployed will need to test creating an extension as well as make sure the correct org names are displayed for users.


edmondsgarrett (approved):
LGTM

PR #94, opened 2023-02-07
hotfix/registration-form-fix

shayanaijaz:
## Overview

Fix for bug in registration form

## Related

…

## Changes

- Bug fix where string value was getting set to `None` instead of `''`

## Testing

- No local testing. After deployment can test by submitting form


edmondsgarrett (approved):
LGTM, good catch!

fnets (approved):
LGTM

PR #97, opened 2023-02-08
Hotfix/fp 1815 exception submission error fix

sophia-massie:
## Overview

- Hotfix to stop submission error of  other exception form
- Hotfix to stop submission error on extension form
- Fix requestor email not saving
- Fix set timestamp to stop saving to midnight for each submission

## Related

- [FP-1815](https://jira.tacc.utexas.edu/browse/FP-1815)
- requires https://github.com/TACC/Core-CMS-Custom/pull/72 Follow on PR from this previous PR

## Changes

- Change cursor error in other form, 
- Make status show as pending not outcome
- Gets requestor email to post to DB

## Notes
TODO
- [ ] Make fields unable to submit if only white space
- [ ] Debug why the _clean_email util when handling _clean_email(form['requestor_email'}) returns None
## Testing

Use snippet on line 32 and 94 of exception/views.py 

submitters = [(1, 'TESTGOLD', 10000000, 'thbrown', 'chcd'), (4, 'TESTMRTN', 10000003, 'thbrown', 'chcd')]

## UI
Threshold Page 1/2
Screen Shot 2023-02-06 at 5 12 10 PM

Threshold page 2/2
Screen Shot 2023-02-06 at 5 14 24 PM

Other exception page 1/2
Screen Shot 2023-02-06 at 5 15 01 PM

Other exception page 2/2
Screen Shot 2023-02-06 at 5 15 16 PM


jmartinez-tacc (approved):
LGTM!

apcd-cms/src/apps/utils/apcd_database.py
```diff
@@ -660,6 +660,7 @@ def create_other_exception(form, sub_data):
     cur = None
     conn = None
     values = ()
+    values = ()
```
shayanaijaz, line 663: Possible duplicate line

sophia-massie: Ah yes I think that was a hold over from a merge. Thank you!

edmondsgarrett (approved):
LGTM, conditionally approved if Shayan's comment is resolved

sophia-massie:
> LGTM, conditionally approved if Shayan's comment is resolved

@edmondsgarrett Done!

shayanaijaz (approved):
LGTM

apcd-cms/src/apps/utils/apcd_database.py
```diff
@@ -690,14 +689,14 @@ def create_other_exception(form, sub_data):
             sub_data[2],
             sub_data[3],
             _clean_value(form['requestor-name']),
-            _clean_email(form['requestor-email']),
+            form['requestor-email'],
```
fnets, line 692: I think the more secure way to solve this problem would be to prevent users from submitting emails that will fail the clean check in the form, similar to how we prevent users from submitting unformatted phone numbers.

apcd-cms/src/apps/utils/apcd_database.py
```diff
@@ -749,15 +747,15 @@ def create_threshold_exception(form, sub_data):
...
```
fnets, line 750: Same as what I referred to in my previous comment about form validation.

fnets (changes requested):
Instead of getting rid of form validation, we should make sure that our form prevents submission of bad data.

apcd-cms/src/apps/utils/apcd_database.py
sophia-massie: So the clean value function when texted will remove everything before the @ sign. I think this needs its own custom util

sophia-massie: Put a new function to clean email and tested on DB. Looks like it works and posts.

sophia-massie (commented):
Added a clean email function that I tested on VM that stores email correctly

fnets (approved):
Instead of changing our backend sanitization to be more accepting, I was thinking that we needed a client-side basic check to make sure that users input at least basically correctly-formed emails. However, this will work for now, and I've put in a PR draft to fix this issue.

PR #103, opened 2023-02-09
View users

davidmtacc:
## Overview
Removed what should now be redundant script as query should now order dropdown list. Also, optimized filter to ignore commas in search. Need to do the same for ampersand but should still work better.
…

## Related

…

## Changes
removed redundant script.. optimized filter script
…

## Testing

Tested using fake data

       user_content = [
            (
                6,
                4,
                'aw@heck.org',
                'Jon Heck',
                'UHC, something',
                'no create',
                'none update',
                'no notes',
                't',
                'submitter'

            ),
            (
                5,
                4,
                'aw@heck.org',
                'Jon Heck',
                'tundra',
                'no create',
                'none update',
                'no notes',
                't',
                'submitter'

            ),
             (
                 5,
                 4,
                'aw@heck.org',
                'Jon Heck',
                'UHC Medical',
                'no create',
                'none update',
                'no notes',
                't',
                'submitter'
            ),           
          ]

## UI
need to work on having table look better in smaller window
…


apcd-cms/src/apps/view_users/views.py
```diff
@@ -16,81 +16,24 @@ def dispatch(self, request, *args, **kwargs):
            return HttpResponseRedirect('/')
         return super(ViewUsersTable, self).dispatch(request, *args, **kwargs)
 
-    def get_context_data(self, *args, **kwargs):
+    def get_context_data(self, user_content, *args, **kwargs):
```
sophia-massie, line 19:
```suggestion
    def get_context_data(self, user_content=user_content, *args, **kwargs):
```

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -41,38 +40,42 @@ <h1>View Users</h1>
...
-                    {{r.notes}}
-                  </td>
+                  <td>{{r.notes}}</td>
               </tr>
           {% endfor %}
       </tbody>
   </table>
 </div>
 <script>
+/*filters by organization with dropdown*/
```
sophia-massie, line 54: UnitedHealthcare (UHC) not pulling up when selecting from the dropdown "UnitedHealthcare (UHC)". UnitedHealthcare (UHC) only shows up under UHC which might have something to do with there being no spaces between UnitedHealthcare ??

sophia-massie (changes requested):

apcd-cms/src/apps/view_users/views.py
davidmtacc: added this to update. Thanks!

apcd-cms/src/apps/view_users/templates/view_users.html
davidmtacc: Hmm. Weird. Current page shows the UnitedHealthcare (UHC) when dropdown chosen. Shouldn't have changed anything there.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -53,15 +53,19 @@ <h1>View Users</h1>
 <script>
 /*filters by organization with dropdown*/
 function filterTableByOrganization() {
-  var dropdown, filter, table, tr, td, i;
+  var dropdown, filter, newinput, bestinput, table, tr, td, i;
   input = document.getElementById("organizationFilter");
-  filter = input.value.toUpperCase();
+  newinput = input.value.replace("&","");
+  bestinput = newinput.replace(",","");
+  console.log(bestinput)
```
edmondsgarrett, line 60:
```suggestion
```

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -53,15 +53,19 @@ <h1>View Users</h1>
...
+  filter = bestinput.toUpperCase();
+  console.log(filter)
```
edmondsgarrett, line 62:
```suggestion
```

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -53,15 +53,19 @@ <h1>View Users</h1>
 <script>
 /*filters by organization with dropdown*/
 function filterTableByOrganization() {
-  var dropdown, filter, table, tr, td, i;
+  var dropdown, filter, newinput, bestinput, table, tr, td, i;
   input = document.getElementById("organizationFilter");
-  filter = input.value.toUpperCase();
+  newinput = input.value.replace("&","");
```
edmondsgarrett, line 58: I wonder if here, we can preprocess the org_names for the filter in the view for this page (to eliminate duplicates, extra characters, etc) so we can just leave str comparison for this function. I only suggest because I see on the staging page that there are some orgs in the dropdown that are the same, just different cases in their names (Ie the org name is both camel-cased and in all lowercase, but both are in the dropdown). What do you think?

edmondsgarrett (changes requested):
Working better for sure! Have a few thoughts and edits.

apcd-cms/src/apps/view_users/templates/view_users.html
sophia-massie: Will have this as a follow on task as per 2.10 scrum

fnets (approved):
LGTM

shayanaijaz (approved):
LGTM

sophia-massie (approved):

PR #110, opened 2023-02-10
fix: Bad default arg for reg id util functions

edmondsgarrett:
## Overview
Needed to change how `WHERE` clause in `get_registrations` & related util functions is included/excluded in SQL command
…

## Related

…

## Changes
- Reverted to f-string solution to completely exclude `WHERE` from SQL command if `reg_id` is not passed
…

## Testing

1. Nothing to test locally.

## UI

…


fnets (approved):
Lgtm

sophia-massie (approved):

shayanaijaz (approved):
LGTM

PR #111, opened 2023-02-13
fix: Eligibility/enrollment file type not storing correctly to database

edmondsgarrett:
## Overview
Added fix for eligibility/enrollment file type not writing to db unless other file types selected
…

## Related

…

## Changes
- In `create_registration` db function, I set `file_me` to always be set as `True` as this file type is mandatory for all registration requests
…

## Testing

1. No testing to do locally.

## UI

…


sophia-massie (approved):
LGTM!

fnets (approved):
LGTM, but don't forget to remove the template mentions `types_of_files_eligibility_enrollment` if it turns out they're not needed.

PR #118, opened 2023-02-23
task/FP-1986: Check Submitter Role

rstijerina:
## Overview

Add `/submissions/check-submitter-role/` endpoint, which provides boolean response based on whether or not the active session user has a CMS & APCD DB role corresponding to a submitter.

Used by the core portal Submissions view.

## Testing

1. https://apcd-qa.tacc.utexas.edu/submissions/submit-file/


apcd-cms/src/apps/submissions/views.py
```diff
@@ -1,12 +1,14 @@
-from django.http import HttpResponseRedirect
+from django.http import HttpResponseRedirect, JsonResponse
 from django.views.generic.base import TemplateView
+from django.contrib.auth.decorators import login_required
 from apps.utils.apcd_database import get_submissions, get_submission_logs
-from apps.utils.apcd_groups import has_apcd_group, is_apcd_admin
+from apps.utils.apcd_groups import has_apcd_group
```
rstijerina, line 5: Lines 5-57 changes are linting-only

sophia-massie (approved):
Checked all roles (including no roles) and the endpoint had the correct response. LGTM

PR #124, opened 2023-03-03
fix: Create submitter form writing bug

edmondsgarrett:
## Overview
Fix for list bug that causes create submitter action to fail on every attempt
…

## Related

…

## Changes
- Added `[0]` index on line in POST which grabs single registration record based on registration id, to correctly index over record's tuple
…

## Testing

1. Nothing to test locally.

## UI

…


sophia-massie (approved):
LGTM!

fnets (approved):
LGTM, and I added a small change to make all the variable names simpler. Tested it on the VM, and works there.

PR #134, opened 2023-03-21
changed column number to reflect change in table and show rolenames

davidmtacc:
## Overview
Addition of 'user-number' to table led to column mismatch. in consultation with Carrie have created a fix to have table show role_names again

## Related

…

## Changes
Changed  column number assigned to role_names from (9) to (10) as user_number was inserted before it.
…

## Testing

1. Can only test against database 

## UI

…


happycodemonkey (approved):
Change looks good to me!

PR #137, opened 2023-03-22
task/FP-1985: Util function for checking inputs for whitespace only

edmondsgarrett:
## Overview
…

## Related

- [FP-1985](https://jira.tacc.utexas.edu/browse/FP-1985)
- [FP-1974](https://jira.tacc.utexas.edu/browse/FP-1974)
- [FP-1976](https://jira.tacc.utexas.edu/browse/FP-1976)
…

## Changes
- Added `.js` script for checking an `<input>` for spaces only, and resetting said `<input>`'s value to nothing so the tag's `required` prop can catch it
- Registered `utils` as custom app for above JS script
- Implemented whitespace check on registration form
…

## Testing

1. Load http://localhost:8000/register/request-to-submit/
2. Fill out form as normal, but fill some text-only inputs with spaces, and try to submit form
3. Confirm form does not successfully submit, and instead redirects back to first field filled with spaces & that said field is now blank (ie with <i>no</i> spaces)
4. Continue trying to submit (depending on number of fields you filled with spaces only) to ensure form catches every instance
5. Now, add at least one additional entity and contact and test whitespace check as before.
6. Also, confirm that, if input with spaces in an additional contact/entity is caught, removing said entity/contact does not block form submission.
7. Confirm that, once all required fields are not whitespace, that form submits.

## UI

…

## Notes
This PR <i>only</i> addresses FP-1985; it contains code to be used in FP-1974 & FP-1976 (apologies for poorly-named branch)
…

### DIFF
--- a/apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
+++ b/apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
@@ -1,0 +1,6 @@
+const formTextInputs = document.querySelectorAll('input[type=text]');
+formTextInputs.forEach(i => i.addEventListener('change', () => {
+    if (i.value.replace(/ /g, '') === '') { 
+      i.value = ''; // if input is all spaces, want to clear it out to leverage <input>'s required prop
+    }
+}))
--- a/apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
+++ b/apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
@@ -1,6 +1,7 @@
-const formTextInputs = document.querySelectorAll('input[type=text]');
-formTextInputs.forEach(i => i.addEventListener('change', () => {
-    if (i.value.replace(/ /g, '') === '') { 
-      i.value = ''; // if input is all spaces, want to clear it out to leverage <input>'s required prop
-    }
-}))
+function noEmptyInputs(inputArr) {
+  inputArr.forEach(i => i.addEventListener('change', () => {
+      if (i.value.replace(/ /g, '') === '') { 
+        i.value = ''; // if input is all spaces, want to clear it out to leverage <input>'s required prop
+      }
+  }))
+}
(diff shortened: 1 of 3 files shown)


apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
```diff
@@ -0,0 +1,6 @@
+const formTextInputs = document.querySelectorAll('input[type=text]');
+formTextInputs.forEach(i => i.addEventListener('change', () => {
```
sophia-massie, line 2: Need to add something here that also checks the text fields within the hidden html for entities and contacts when more than one is added. It's letting me submit with blank second inputs in the second contact and second entity.

sophia-massie (changes requested):
This is amazing and will be so helpful for all forms on this project. Thank you so much! Only thing is it's missing a catch for the additional contacts and entities on the register to submit form.

sophia-massie (approved):
LGTM and works locally! Thanks for the quick turn around on this it's a huge help!

apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
fnets, line 1: Is there a reason this wasn't put into `utils/utils.py`? Then you could clean up a lot of the other files, and also make that file a little more useful.

edmondsgarrett: Oh yeah, well I first created a separate script for this as its JavaScript, and then I placed it where I did so Django would pick it up and I could load it into an HTML template via static

jmartinez-tacc (approved):
LGTM!

apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
fnets: Duh! Didn't even realize what language I was reading at the end of the day!

fnets (approved):

PR #141, opened 2023-03-30
fix(apcd-cms): add+load css not in core-styles v2 ⏸️

wesleyboar:
## Overview

Limit input field length (because [TACC/Core-Styles] does not do it in newer version).

<details>

[TACC/Core-Styles](https://github.com/TACC/Core-Styles) does not limit input field length (as of its [e068073]). But APCD relies on it (the limit was built for APCD's first form.) This commit need not be in main until APCD loads Core-CMS after merge of https://github.com/TACC/Core-CMS/pull/569.

</details>

## Related

- prepares for https://github.com/TACC/Core-Styles/commit/e068073
    - APCD would **not** have Core-Styles v2 **until** it sets `TACC_CORE_STYLES_VERSION = 2`.
    - APCD setting `TACC_CORE_STYLES_VERSION = 2` would have **no** effect **until** merge of [#159].

[#159]: https://github.com/TACC/Core-CMS-Custom/pull/159

## Changes

- **added** `assets_custom.html`
- **added** `site.cms.css`

## Testing

0. Use browser on wide screen.
1. Open https://apcd-qa.tacc.utexas.edu/register/request-to-submit/ (or local equivalent).
2. Verify form fields do not exceed 50 characters.

## UI

…

## Notes

The `assets_custom.html` is a template that Core-CMS has and automatically loads. The code, https://github.com/TACC/Core-CMS-Custom/blob/2b78220/apcd-cms/Dockerfile#L6, should cause it to override the one in CMS. Untested.

[TACC/Core-Styles]: https://github.com/TACC/Core-Styles
[e068073]: https://github.com/TACC/Core-Styles/commit/e068073


wesleyboar:
This PR still seems relevant, even after #159. Unless you add comment showing you verify otherwise, please do not close this PR. Failing to implement this with #159 might cause a form field width bug. In fact, maybe I merge this with #159… I'll check later.

wesleyboar:
**Update**: This is **not** rushed. It actually adds styles that are redundant _at the moment_, **but** prevents an input field width bug when APCD `TACC_CORE_STYLES_VERSION = 2` (unscheduled)—by then I'll forget about this PR. So I am merging now.

PR #145, opened 2023-04-12
task/WP-19: VIew users filtering in view rather than in template

edmondsgarrett:
## Overview
Moved view users data filtering to view to accommodate upcoming pagination on page
…

## Related

- [WP-19](https://jira.tacc.utexas.edu/browse/WP-19)
- [WP-10](https://jira.tacc.utexas.edu/browse/WP-10)
…

## Changes
- Pared down filter function in template to grab selected filter and pass back to view via XML request
- Added new user object value to remove parentheses from organization names (replicate filter check already happening on this page)
- Filtering data in view based on url parameter passed from template
- Moved filter options to view for setting up dropdown option as currently selected filter
- Added url pattern to `view-users/urls.py` which accommodates filter url parameter
- Edited filter util function to check for filter substring in queried category data 
…

## Testing

1. Replace line 12 of `view_users/views.py` with the following:
   <details><summary>Snippet</summary>
     
         user_content = [
        (
            6,
            4,
            '[aw@heck.org](mailto:aw@heck.org)',
            'Jon Heck',
            'UHC & Nop',
            'no create',
            'none update',
            'no notes',
            'True',
            7,
            'submitter'
        ),
        (
            5,
            4,
            '[aw@heck.org](mailto:aw@heck.org)',
            'Alex Hamilton',
            't and (t)',
            'no create',
            'none update',
            'no notes',
            'True',
            8,
            'submitter',
        ),
        (
            5,
            4,
            '[aw@heck.org](mailto:aw@heck.org)',
            'Penny Ante',
            'UHC',
            'no create',
            'none update',
            'no notes',
            'True',
            10,
            'submitter'
        ),
        ]
        
    </details>
2. Load http://localhost:8000/administration/view-users/
3. Click through the options in the filter dropdown and ensure:
    - Page is reloaded on filter selection, and table only shows entries where the organization matches (or partially matches) the organization on the page
    - Filter dropdown shows the filter you selected as the default option
 

## UI

…

### DIFF
--- a/apcd-cms/src/apps/view_users/templates/view_users.html
+++ b/apcd-cms/src/apps/view_users/templates/view_users.html
@@ -18,9 +18,8 @@
     <div class="filter-content">
       <span><b>View Users by Organization: </b></span>
       <select id="organizationFilter" class="organization-filter" onchange="filterTableByOrganization()">
-        <option class="dropdown-text" value="">All</option>
-        {% for r in rows %}
-        <option class="dropdown-text">{{r.org_name}}</option>
+        {% for option in filter_options %}
+          <option class="dropdown-text" {% if option == selected_filter %}selected{% endif %}>{{ option }}</option>
         {% endfor %}
       </select>
     </div>
@@ -54,21 +53,14 @@
 <script>
 /*filters by organization with dropdown*/
 function filterTableByOrganization() {
-  var dropdown, filter, newinput, bestinput, table, tr, td, i;
+  var filter, xhr;
   input = document.getElementById("organizationFilter");
-  filter = input.value.replace("&", "&amp;").replace("(", "").replace(")","");
-  table = document.getElementById("usersTableBody");
-  tr = table.getElementsByTagName("tr");
-  for (i = 0; i < tr.length; i++) {
-    td = tr[i].getElementsByTagName("td")[2];
-    if (td) {
-      if (td.innerHTML.replace("(", "").replace(")","").search(filter) > -1) {
-        tr[i].style.display = "";
-      } else {
-        tr[i].style.display = "none";  
-      }
-    }  
-  }
+  filter = input.value.replace("&", encodeURIComponent('&'));
+  xhr = new XMLHttpRequest();
+  xhr.open('GET', `/administration/view-users/?filter=${filter}`);
+  xhr.send();
+  window.location.href = `/administration/view-users/?filter=${filter}`;
+  window.location.load();
 }
 
 /* remove duplicates from dropdown */
--- a/apcd-cms/src/apps/view_users/urls.py
+++ b/apcd-cms/src/apps/view_users/urls.py
@@ -1,4 +1,4 @@
-from django.urls import path, re_path
+from django.urls import path
 from apps.view_users.views import ViewUsersTable
 
 app_name = 'administration'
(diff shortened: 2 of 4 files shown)


apcd-cms/src/apps/view_users/urls.py
```diff
@@ -1,7 +1,8 @@
-from django.urls import path
+from django.urls import path, re_path
 from apps.view_users.views import ViewUsersTable
 
 app_name = 'administration'
 urlpatterns = [
     path('view-users/', ViewUsersTable.as_view()),
+    re_path(r'view-users/?filter=(?P<filter>[\w\ ]+)/', ViewUsersTable.as_view()),
```
sophia-massie, line 7:
```suggestion
    path('view-users/<str:filter>/', ViewUsersTable.as_view())
```

Just a suggestion - feel free to ignore. Would there be any instance where the filter will not be passed as a string? If not, you could use this pattern to be a bit more easier to read.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -54,21 +53,14 @@ <h1>View Users</h1>
...
-    td = tr[i].getElementsByTagName("td")[2];
-    if (td) {
-      if (td.innerHTML.replace("(", "").replace(")","").search(filter) > -1) {
-        tr[i].style.display = "";
-      } else {
-        tr[i].style.display = "none";  
-      }
-    }  
-  }
+  filter = input.value.replace("&", encodeURIComponent('&'));
```
sophia-massie, line 58: It may make sense to eventually switch this over to an AJAX call so the view can get the filter information without reloading the page. 
I did something similar for the exception threshold [here ](https://github.com/TACC/Core-CMS-Custom/blob/0e76a393f976a783cbe97603e809893ebb8147e3/apcd-cms/src/apps/exception/templates/exception_submission_form/exception_threshold_form.html#L257). This may require follow on work to update functions in your view to get the filter param and your URL patterns. So this is just a suggestion for down the line.

sophia-massie (approved):

apcd-cms/src/apps/view_users/urls.py
sophia-massie: I tested this locally and it has the same functionality as the original PR

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -54,21 +53,14 @@ <h1>View Users</h1>
...
+  xhr = new XMLHttpRequest();
```
sophia-massie, line 59: It may make sense to eventually switch this over to an AJAX call so the view can get the filter information without reloading the page.
I did something similar for the exception threshold here . This may require follow on work to update functions in your view to get the filter param and your URL patterns. So this is just a suggestion for down the line.

apcd-cms/src/apps/view_users/urls.py
```diff
@@ -1,7 +1,8 @@
-from django.urls import path
+from django.urls import path, re_path
```
sophia-massie, line 1:
```suggestion
from django.urls import path
```

jmartinez-tacc (approved):
LGTM!

apcd-cms/src/apps/view_users/urls.py
sophia-massie: Only if you end up going with the path <str:filter> solution

apcd-cms/src/apps/view_users/templates/view_users.html
edmondsgarrett: Good point! I'll make a Jira task for this.

apcd-cms/src/apps/view_users/urls.py
edmondsgarrett: Yeah, this makes sense! Committing.

PR #154, opened 2023-05-09
fix: Add correct codes for org type to reg form template

edmondsgarrett:
## Overview

…

## Related

…

## Changes
- On registration form (for new registration, not edit), modified values of `option` elements to match codes from standard codes table
…

## Testing

1. Nothing to test locally; confirmed on QA and standard codes table that these codes are correct

## UI

…


fnets (approved):
lgtm

sophia-massie (approved):
LGTM!

PR #160, opened 2023-05-23
Create status filter/jira wp 9 user listing filter on status

davidmtacc:
## Overview
Added filter for status (active/inactive) plagiarizing Garrett's work shamlessly. I also added some script that shows text that no data available when filters result in an empty table
…

## Related

…

## Changes
I added an additional dropdown that show status states and added the button that clears filters. I also added script checking for an empty table that shows notification if no content.
…

## Testing

1. I tested using data below
        user_content = [
            (
                6,
                4,
                'Rob@arby.org',
                'Robert Smith',
                'UGT',
                '10/22/33',
                '11/22/23',
                'notes here',
                True,
                7,
                'submitter'
        
            ),
            (
                6,
                4,
                'simon@arby.org',
                'Simon Gallup',
                'Dunstreet',
                '10/22/33',
                '11/22/23',
                'notes here',
                False,
                2,
                'admin'
        
            ),
            (
                6,
                4,
                'rog@arby.org',
                'Roger ODonnel',
                'Dunstreet',
                '10/22/33',
                '11/22/23',
                'notes here',
                True,
                3,
                'submitter'
        
            ),
            (
                6,
                4,
                'jace@arby.org',
                'Jason Cooper',
                'UGT',
                '10/22/33',
                '11/22/23',
                'notes here',
                False,
                6,
                'admin'
        
            ),
            (
                6,
                4,
                'rere@arby.org',
                'Reeves Gabrels',
                'P',
                '10/22/33',
                '11/22/23',
                'notes here',
                False,
                5,
                'admin'
        
            ),
        ]

## UI

…Using test data I checked if the correct content shows up when various filter combinations are chosen. Table reads clearly in all. 
Screenshot 2023-05-23 at 7 29 41 AM


apcd-cms/src/apps/view_users/views.py
```diff
@@ -19,22 +19,6 @@ def dispatch(self, request, *args, **kwargs):
     def get_context_data(self, user_content=user_content, *args, **kwargs):
         context = super(ViewUsersTable, self).get_context_data(*args, **kwargs)
 
-        def _set_user(usr):
```
edmondsgarrett, line 22: Why was this function removed but not its uses?

edmondsgarrett (changes requested):
With testing, I saw that this page doesn't work with the current PR, needs to be updated. See my comment below.

apcd-cms/src/apps/view_users/views.py
davidmtacc: I was a bit confused myself. But I'm pretty sure I haven't deleted anything that should be here, this time. Thank you for testing again if you get a chance.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -14,16 +14,25 @@ <h1>View Users</h1>
...
     <div class="filter-content">
       <span><b>View Users by Organization: </b></span>
       <select id="organizationFilter" class="organization-filter" onchange="filterTableByOrganization()">
         {% for option in filter_options %}
-          <option class="dropdown-text" {% if option == selected_filter %}selected{% endif %}>{{ option }}</option>
+          <option class="dropdown-text" {% if option == selected_org %}selected{% endif %}>{{ option }}</option>
         {% endfor %}
       </select>
     </div>
+    <select id="statusFilter" onchange="filterTableByStatus()">
```
edmondsgarrett, line 26: This drop down will need a label, like the organization filter above.

edmondsgarrett (changes requested):
Choosing an option in the drop down successfully triggers page reload with relevant data only on page, nice!

A couple of things I noticed:
- Filter does not show selected option, just shows 'All' no matter what choice is made; I think you can copy what is done with the org filter on this page to get the selected option to show on the drop down.
- Looks like some of the style reworks Wes contributed need to be merged in to this branch, which will require the merge conflict on this branch to be resolved.
- One other note made below.

apcd-cms/src/apps/view_users/templates/view_users.html
davidmtacc: Cool. Thanks for the feedback.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -14,22 +14,35 @@ <h1>View Users</h1>
   <hr />
   <p style="margin-bottom: 30px">View submitted users.</p>
   <hr />
-  <div class="filter-container">
+  <div class="filter-container" style="display: flex; gap: 10px;">
     <div class="filter-content">
       <span><b>View Users by Organization: </b></span>
       <select id="organizationFilter" class="organization-filter" onchange="filterTableByOrganization()">
```
edmondsgarrett, line 20:
```suggestion
      <select id="organizationFilter" class="status-filter" onchange="filterTableByOrganization()">
```
Looks like the class `organization-filter` doesn't exist, this class should work for matching other filters on the website.

apcd-cms/src/apps/view_users/views.py
```diff
@@ -30,29 +31,46 @@ def _set_user(usr):
                     'created_at': usr[5],
                     'updated_at': usr[6],
                     'notes': usr[7],
-                    'active': 'Active' if usr[8] else 'Inactive',
+                    'status': 'Active' if usr[8] else 'Inactive',
                     'user_number': usr[9],
                     'role_name': usr[10],
                     'org_name_no_parens': usr[4].replace("(", "").replace(")", ""),  # just for filtering purposes
+                    'active': 'Yes' if usr[8] else 'No', # just to tell active from inactive status
```
edmondsgarrett, line 38: I'm unsure if this necessary, I think you could use the `status` field and display Active or Inactive in the filter instead. I'm curious what others think!

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -14,22 +14,35 @@ <h1>View Users</h1>
...
         {% for option in filter_options %}
-          <option class="dropdown-text" {% if option == selected_filter %}selected{% endif %}>{{ option }}</option>
+          <option class="dropdown-text" {% if option == selected_org %}selected{% endif %}>{{ option }}</option>
         {% endfor %}
       </select>
     </div>
+    <div class="filter-content">
```
edmondsgarrett, line 26:
```suggestion
```
To match other pages, you can wrap all of the filters & clear button in the `filter-content` div.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -14,22 +14,35 @@ <h1>View Users</h1>
...
+      <span><b>Is Status Active: </b></span>
+      <select id="statusFilter" class="status-filter" onchange="filterTableByStatus()">
+        {% for option in status_options %}
+        <option class="dropdown-text" {% if option == selected_status %}selected{% endif %}>{{ option }}</option>
+      {% endfor %}
+      </select>
+   </div>
```
edmondsgarrett, line 33:
```suggestion
```
Related to `filter-content` comment.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -14,22 +14,35 @@ <h1>View Users</h1>
...
+  {% if selected_status or selected_org %}
+        <button onclick="clearSelections()">Clear Options</button>
+      {% endif %}
   </div>
```
edmondsgarrett, line 37:
```suggestion
     </div>
  </div>
```
Related to `filter-content` comment above.

apcd-cms/src/apps/view_users/templates/view_users.html
```diff
@@ -14,22 +14,35 @@ <h1>View Users</h1>
   <hr />
   <p style="margin-bottom: 30px">View submitted users.</p>
   <hr />
-  <div class="filter-container">
+  <div class="filter-container" style="display: flex; gap: 10px;">
```
edmondsgarrett, line 17:
```suggestion
  <div class="filter-container">
```
Related to `filter-content` comment, becomes unneeded.

edmondsgarrett (changes requested):
Have some styling comments, looks good otherwise!

apcd-cms/src/apps/view_users/views.py
davidmtacc: Yeh. I did what you suggested first but, the filter couldn't differentiate between 'ACTIVE' and 'inACTIVE' because it kept matcing.. ACTIVE

edmondsgarrett: Ah that makes sense - that seems buggy for a table filter. I'll add a change for that.

edmondsgarrett (approved):
LGTM, nice work!

sophia-massie (approved):
LGTM!

PR #178, opened 2023-06-28
bug/WP-181: Entity codes not saving on edit action

edmondsgarrett:
## Overview
Addresses bug where edits on an entity in 'Edit Registration' action were not written to database despite success message.
…

## Related

- [WP-181](https://jira.tacc.utexas.edu/browse/WP-181)
…

## Changes
- On edit registration template, edited entity form element names to differentiate from records on the same listing page (was causing some fields to not show on form at all/not be found when writing to database, causing silent failure)
- Updated action for updating entities via above form to look for updated element names
…

## Testing

2. On any record, open the 'Edit Record' option in the 'Actions' column drop down.
3. Make some edits to any of the fields in the entity section (some records here will not have entities (growing pains with local db testing), skip those as not representative of real data) and submit changes, ensure you are directed to a success page.
4. Navigate back to listing page and either View or Edit the same record and ensure changes you made now appear.

## UI

…


apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
```diff
@@ -268,14 +269,14 @@ <h5 id="entity_header_{{ forloop.counter }}_{{r.reg_id}}">Entity {{ forloop.coun
                 <div class="field-wrapper textinput">
                     <div class="field-errors" style="display: none"></div>
 
-                    <label for="fein_{{forloop.counter}}_{{r.reg_id}}"> FEIN<sup>5</sup> </label>
+                    <label for="fein_{{ent_no}}_{{r.reg_id}}"> FEIN<sup>5</sup> </label>
```
sophia-massie, line 272: The modal may need access to the JS that allows only one of the Number/Codes to be filled out (FEIN, License, or company code). The modal is requiring all three to be filled out.

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
```diff
@@ -345,25 +346,25 @@ <h6>File Submission</h6>
...
+                        {% for file_type, file_type_selected in entity.files_type.items %}
                         {% if not forloop.first %}
                         <li>
-                            <label for="types_of_files_{{ file_type|lower }}_{{forloop.counter}}_{{r.reg_id}}"
+                            <label for="types_of_files_{{ file_type|lower }}_{{ent_no}}_{{r.reg_id}}"
                                 ><input
                                 type="checkbox"
-                                name="types_of_files_{{ file_type|lower }}_{{forloop.counter}}_{{r.reg_id}}"
-                                id="types_of_files_{{ file_type|lower }}_{{forloop.counter}}_{{r.reg_id}}"
+                                name="types_of_files_{{ file_type|lower }}_{{ent_no}}_{{r.reg_id}}"
```
sophia-massie, line 366: Should this file type field be required for each entity? Ignore this comment if not.

apcd-cms/src/apps/utils/apcd_database.py
```diff
@@ -328,13 +328,13 @@ def update_registration_entity(form, reg_id, iteration, no_entities):
     conn = None
     values = ()
     try:
+        str_end = f'{iteration}_{reg_id}'
         if not _acceptable_entity(form, iteration, reg_id):
             if iteration <= no_entities: # entity is not in form but was in original entity list -> need to delete
-                return delete_registration_entity(reg_id, form[f'ent_id_{iteration}'])
+                return delete_registration_entity(reg_id, form['ent_id_{}'.format(str_end)])
```
sophia-massie, line 334: When I hit the remove entity button, I get a console error. It only seems to remove the Entity 2 header, but doesn't remove the last entity. If you exit modal and come back, the entity and is still there and the Entity 2 header comes back. Is this button supposed to remove that entity from the DB for this registered submitter?
```
Uncaught TypeError: Cannot set properties of null (setting 'innerHTML')
    at HTMLButtonElement.<anonymous> 
```

sophia-massie (changes requested):
Looks like the entity is saving now! Looks great!  A couple of issues:

-  the remove entity button only removes the Entity header and not the entity itself
-  the Number/Codes fields (FEIN and NAIC) numbers require all three to be filled out for each entity in order to submit
- The file submission type is not required and you can submit without selecting. Is this intentional? If so maybe remove the asterisk so it no longer says required.

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
edmondsgarrett: I'll take the (required) part out of this section, but UTH does want those checkboxes on each entity (documenting our slack conversation about this here for relevance).

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
edmondsgarrett: Should be fixed now!

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
edmondsgarrett: Done.

apcd-cms/src/apps/utils/apcd_database.py
edmondsgarrett: Fixed!

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_scripts.html
```diff
@@ -58,9 +58,17 @@
         {% else %}
             1
         {% endif %};
+      let reg_id = 
+        {% if r %}
+          Number(`
+            {{r.reg_id}}
```
fnets, line 64: Can you combine 63-65 into one line? It seems a little more readable to me, but if it's just splitting hairs ignore me.

fnets (approved):

PR #190, opened 2023-08-08
Sorting by incorrect date field for exceptions and extensions

sophia-massie:
## Overview

Fixing a bug where the exceptions and extensions were not sorting by created date when no filters were selected

## Related

- [WP-218](https://jira.tacc.utexas.edu/browse/WP-218)

## Changes

Updated view that shows admin exception and extension records so that it is sorted by the correct field in the database (created_at)

## Testing

1. Go to [http://localhost:8000/administration/list-exceptions/](http://localhost:8000/administration/list-exceptions/ )
2. Make sure they're listed in order from created at column. 
3. Create a new exception and make sure it is the first on the admin list
4. Go to [http://localhost:8000/administration/list-extensions/](http://localhost:8000/administration/list-extensions/and) select Edit Record from Action drop down
2. Make sure they're listed in order from created at column. 
3. Create a new extension and make sure it is the first on the admin list

### DIFF
--- a/apcd-cms/src/apps/admin_exception/views.py
+++ b/apcd-cms/src/apps/admin_exception/views.py
@@ -97,7 +97,7 @@ class AdminExceptionsTable(TemplateView):
         context['exceptions'] = []
 
         def getDate(row):
-            date = row[1]
+            date = row[6]
             return date if date is not None else parser.parse('1-1-0001')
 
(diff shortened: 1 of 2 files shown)


happycodemonkey (approved):
Looks good!

edmondsgarrett (approved):
LGTM!

PR #195, opened 2023-08-22
hotfix/WP-232-234-APCD-frontend-form-sanitation

sophia-massie:
## Overview
The purpose of this hotfix is to still allow special characters that users may use but restricting invalid characters on form submissions. 
Your thoroughness of testing is really appreciated! I know it's tedious. Using Django or React utilities in place of this band-aid is the long term goal once the fire is put out.

## Related

- [WP-232](https://jira.tacc.utexas.edu/browse/WP-232)
- [WP-233](https://jira.tacc.utexas.edu/browse/WP-233)
- [WP-234](https://jira.tacc.utexas.edu/browse/WP-234)
- mimicked by #201

## Changes

The checkForBlankInputs javascript was changed using whitelist methods to allow characters in input. At the same time, the script blacklists some explicit scripting character patterns. This is a front end change to prevent bad characters from making it to the server. The DB utils remove bad characters server side in apcd_database.py. 

## Testing

1. Go to [http://localhost:8000/register/request-to-submit/](http://localhost:8000/register/request-to-submit/)
2. Try to input **<script> </script> javascript.exe, anything in "quotes" or 'quotes" { hello.exe } thisshouldwork@test.com into any field** into a field
3. On change. the field should be replaced  with **Try to input script script exe anything in quotes or quotes  helloexe  thisshouldworktestcom**
4. Go to [http://localhost:8000/submissions/threshold-exception/](http://localhost:8000/submissions/threshold-exception/)
5. Repeat steps 2 and 3
6. Go to [http://localhost:8000/submissions/other-exception/](http://localhost:8000/submissions/other-exception/)
7. Repeat steps 2 and 3
8. Go to [https://apcd-qa.tacc.utexas.edu/submissions/extension-request/](https://apcd-qa.tacc.utexas.edu/submissions/extension-request/)
9.  Repeat steps 2 and 3

## UI

###User input:

Screenshot 2023-08-22 at 6 07 51 PM

###On change, text is changed to :

Screenshot 2023-08-22 at 6 08 53 PM


apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
```diff
@@ -1,7 +1,11 @@
 function noEmptyInputs(inputArr) {
   inputArr.forEach(i => i.addEventListener('change', () => {
-      if (i.value.replace(/ /g, '') === '') { 
-        i.value = ''; // if input is all spaces, want to clear it out to leverage <input>'s required prop
-      }
+    const cleanedValue = i.value.replace(/[^a-zA-Z0-9\s-@!?,%$#&*().]|<script>|<\/script>|'|"|}|{|javascript|JavaScript|\.exe/gi, '') // Allow specified characters
```
chandra-tacc, line 3: Have we thought about tools like [DOMPurify](https://github.com/cure53/DOMPurify)? instead of us writing a regex and maintaining this?

sophia-massie: Thanks for working with me this morning on this! Great solution. I will get with @wesleyboar about possibly adding package to Core-CMS. I created the follow on task here. https://jira.tacc.utexas.edu/browse/WP-240

chandra-tacc (approved):
LGTM.
Tested the regex on bad strings. Also, one of the forms worked for me and tested it.
The follow on future task is to use a public package to sanitize instead of regex

apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
wesleyboar: Excellent suggestion. I'll evaluate how it could be put in Core.[^1]

I think you can use it in APCD-CMS as is, without waiting on Core.

<details><summary>Do We Use External JavaScript like that in a TACC CMS?</summary>

_**Yes.** [BrainMap](https://portal.brainmap.org) is one of several sites with [a snippet that loads and uses external JavaScript](https://portal.brainmap.org/admin/djangocms_snippet/snippet/4/change/). The snippet imports JavaScript form an external CDN, then uses it to manipulate the DOM._

```
<script type="module" id="js-ext-link-open-new-tab">
    import findLinksAndSetTargets from 'https://cdn.jsdelivr.net/gh/TACC/Core-CMS@v3.11.0-alpha.1/taccsite_cms/static/site_cms/js/modules/setTargetForExternalLinks.js';

    /* To make certain links open in new window, and apply a11y attribute */
    findLinksAndSetTargets({
        pathsToForceSetTarget: [
            '/tools/sleuth/',
            '/tools/gingerale/',
            '/tools/mango/',
            '/guides/brainmap-cli-hot-guide-redirect/'
        ]
    });
</script>
```

</details>

<details><summary>How Can I Use this on APCD CMS Right Now?</summary>

_I have enough immediate bandwidth only to show you how to **load** it, and **begin to use** it. Please attempt to implement it on APCD CMS as needed. That can teach me how Core can provide it for a client.[^1]_

**Local Usage**

Local Load & Usage
```
<script type="module" id="js-purify-dom">
	import * as DOMPurify from 'dompurify';

	const specificFieldDataA = ''/* ... */;
	const cleanedFieldDataA = DOMPurify.sanitize(specificFieldDataA);

    /* ... */
</script>
```

**Global Usage**

Global Load
```
<script src="dist/purify.min.js"></script>
```

Local Use of Global Load
```
<script>
	const specificFieldDataA = ''/* ... */;
	const cleanedFieldDataA = DOMPurify.sanitize(specificFieldDataA);

    /* ... */
</script>
```

</details>

[^1]: (Warning: Train of Thought) I could make it load on pages that use Django CMS Form, and sanitize all(?) field input blindly, or limit it to only fields with a certain name. But then I must rely on CMS admins to create forms using the "correct" names, but policing that is overhead, and overhead I cannot automate. Attempting to integrate with Django CMS Form also worries me, because that outdated and barely-maintained plugin is one reason we can't quickly update to latest version of Django...

edmondsgarrett (changes requested):
Everything seems to be working, but I noticed that instances of .exe on my test record are making it through on form submit.

apcd-cms/src/apps/utils/static/utils/js/checkForBlankInputs.js
wesleyboar: Edited message. I fixed string typo. Basically, it can be loaded as module from CDN via https://cdn.jsdelivr.net/gh/cure53/DOMPurify@3.0.5/dist/purify.js.

happycodemonkey:
Adding a recommendation to strip out ".sh" extensions as well if we're doing .exe

happycodemonkey (approved):
This looks good to me, tested in both Firefox and Chrome.

wesleyboar (commented):
I approve the idea, and that there is a follow-up to use third-party code.

I trust others' testing. Please check [the bug that E.G. experienced](https://github.com/TACC/Core-CMS-Custom/pull/195#pullrequestreview-1592077151).

Maybe helpful: I tested the regex on `malicious.exe` on three tools. One failed.

| screenshots |
| - |
| test 1 | test 2 | test 3 |

sophia-massie:
> Everything seems to be working, but I noticed that instances of .exe on my test record are making it through on form submit.

Updated to prevent this. All input fields will be checked on these pages now.

wesleyboar (approved):
I had worried the [recent change to select **all** inputs](https://github.com/TACC/Core-CMS-Custom/pull/195#issuecomment-1690540218) would break on `type="checkbox"` or `type="file"`. No problem, because the values of those fields are strings, even if omitted or empty.

<details>

> If the value attribute was omitted, the default value for the checkbox is on, so the submitted data in that case would be subscribe=on.
> — [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value)

> A file input's [value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#value) attribute contains a string that represents the path to the selected file(s). If no file is selected yet, the value is an empty string (""). When the user selected multiple files, the value represents the first file in the list of files they selected.
> — [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#value)

</details>

edmondsgarrett (approved):
LGTM

wesleyboar (commented):
For when you revisit this:
1. The `noEmptyInputs` is no longer an accurate function name; it does more now.
2. Can empty inputs be prevented more simply via [`required` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)?
3. Lots of duplication (but I know you're already investigating a solution).

PR #207, opened 2023-08-28
hotfix/APCD-prod-validation-fix-for-added-entities

sophia-massie:
## Overview

…

## Related

…

## Changes

…

## Testing

1.

## UI

…

### DIFF
--- a/apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
+++ b/apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
@@ -1,4 +1,4 @@
-<form action="" method="POST">
+<form id="registration-form" action="" method="POST">
     {% csrf_token %}
 
     {% if r %}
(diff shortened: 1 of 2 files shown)


apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
```diff
@@ -581,16 +467,16 @@ <h4>
...
                 </label>
 
                 <input
                 type="text"
-                name="entity_name"
+                name="entity_name_1"
                 required
                 class="textinput"
-                id="entity_name"
+                id="entity_name_1"
```
edmondsgarrett, line 479: One example of labels that have changed between staging and prod deployments.

edmondsgarrett (changes requested):
With placing the staging version of these files here, there will also need to be updates to the related database util functions. I pointed to one example of this, and you may want to check the registration table views for the variable names that are used in these templates (i.e. `r.{some var name from view}`).

happycodemonkey (approved):

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_body.html
sophia-massie: I edited all the queries used by the registration app's view so that they match the staging queries. Here are the modified queries. 

- create_registration
- create_registration_entity
- create_registration_contact

apcd-cms/src/apps/registrations/templates/submission_form/registration_form_scripts.html
```diff
@@ -267,7 +365,7 @@ <h6>
                 class="telephoneinput"
                 id="contact_phone_${contacts}{% if r %}_{{r.reg_id}}{% endif %}"
                 inputmode="tel"
-                pattern="^(\\+0?1\\s)?\\(?\\d{3}\\)?[\\s.-]\\d{3}[\\s.-]\\d{4}$"
+                pattern="^(\\+0?1\\s)?\\(?(?![{}'"<>\\/])\\d{3}\\)?(?![{}'"<>\\/])[\\s.-](?![{}'"<>\\/])\\d{3}(?![{}'"<>\\/])[\\s.-](?![{}'"<>\\/])\\d{4}$"
```
sophia-massie, line 368: Had to update the pattern for the telephone number input when adding more than one contact. It was the only field still allowing invalid characters on submission.

Screenshot 2023-08-28 at 5 51 02 PM

edmondsgarrett (approved):
LGTM, nice work!

PR #215, opened 2023-08-30
Remove admin, exception, extension, and submission table from prod

sophia-massie:
## Overview

To remove apps from prod so that only request to submit is allowed 

## Changes

Removed custom urls and apps for pages we do not need.

## Testing

1. Ensure that http://localhost:8000/register/request-to-submit/ still works and has input validators the same as previous PRs for prod
2. Go to each page below and make sure there is no error, but no page loads either. 

- http://localhost:8000/administration/list-registration-requests/
-  http://localhost:8000/submissions/exception/ 
- http://localhost:8000/submissions/extension-request/
- http://localhost:8000/administration/view-users/
- http://localhost:8000/administration/list-exceptions/
- http://localhost:8000/administration/list-extensions/
- http://localhost:8000/administration/list-submissions/

NOTE: There still is communication with the server this way. This is simply making the templates not render at all on the page. This does cause a 404

## UI

…


edmondsgarrett (approved):
LGTM, thanks for the quick work!

happycodemonkey (approved):
all listed links go to 404 as described

PR #229, opened 2023-10-09
Task/wp 312 view users pagination

mjames-uth:
## Overview

This is the first round in integrating the API into the apcd-cms project. This required a custom paginator that could interact with the API.

To use:
1.Pull branch
2. Update the secrets file with the following code: APCD_API_URL = `'http://XX.XX.XX.XX'`
This IP should be to the API making sure there is no forward slash `/` at the end

## Related

…

## Changes

Custom API Paginator - is built on top of the current django paginator. It takes only a subset of data, but still calculates page numbers and size correctly.
ADMIN > View User - The View Users page was used to demonstrator how the paginator works and that it works correctly

## Testing

Confirm all of the same functionality exists on the View Users Page
1. View Users Page Loads
2. Paging Works - Correct Number of Pages
3. Filter by Active/Inactive Status
4. Filter by Orgs
5. View Record
6. Edit Record

## UI

…

### DIFF
--- a/apcd-cms/src/apps/view_users/views.py
+++ b/apcd-cms/src/apps/view_users/views.py
@@ -1,10 +1,8 @@
-from django.http import HttpResponseRedirect, HttpResponse, JsonResponse
-from django.core.paginator import Paginator, EmptyPage
+from django.http import HttpResponseRedirect, HttpResponse
 from django.views.generic.base import TemplateView
 from django.template import loader
-from apps.utils.apcd_database import get_users, update_user, get_api_users, update_api_users, get_api_users_org_list
+from apps.utils.apcd_database import  get_api_users, update_api_users, get_api_users_org_list
 from apps.utils.apcd_groups import is_apcd_admin
-from apps.utils.utils import table_filter
 from apps.components.paginator.custom_api_paginator import paginator
 import logging
 
@@ -52,17 +50,6 @@ class ViewUsersTable(TemplateView):
         template = _edit_user(form)
         return HttpResponse(template.render({}, request))
 
-    # def get(self, request, *args, **kwargs):
-    #     user_content = get_users()
-    #     # print(f"Get API users: {get_api_users()}")
-    #
-    #     context = self.get_context_data(user_content, *args, **kwargs)
-    #     template = loader.get_template(self.template_name)
-    #     return HttpResponse(template.render(context, request))
-    #     ##END FORM FUNCTION
-    #
-    # user_content = get_users()
-
     def dispatch(self, request, *args, **kwargs):
         if not request.user.is_authenticated or not is_apcd_admin(request.user):
             return HttpResponseRedirect('/')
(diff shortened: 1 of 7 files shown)


apcd-cms/src/apps/view_users/views.py
```diff
@@ -36,93 +48,72 @@ def _edit_user(form):
...
-        context['header'] = ['User ID', 'Name', 'Organization', 'Role', 'Status', 'User Number', 'See More']
+        context['header'] = ['User ID', 'Name', 'Organization', 'Role', 'Active', 'User Number', 'See More']
         context['status_options'] = ['All', 'Active', 'Inactive']
-        context['filter_options'] = ['All']
-        context['role_options'] = ['SUBMITTER_USER', 'SUBMITTER_ADMIN','APCD_ADMIN']
-
-        context['status'] = ['Active', 'Inactive']
+        context['role_options'] = ['SUBMITTER_USER', 'SUBMITTER_ADMIN', 'APCD_ADMIN']
+        context['status'] = ['True', 'False']
+        context['filter_options'] = ['All'] + get_api_users_org_list()
```
sophia-massie, line 78: This is throwing an error 'can only concatenate list (not "dict") to list'.

You could try something like 
` org_list = get_api_users_org_list()
        for i in org_list:
            context['filter_options'].append(i)`

sophia-massie (changes requested):
Appending the org_filters to the context['filter_options'] should address the initial error. Otherwise, looks great to me!

sophia-massie (approved):
Looks great to me!

sophia-massie (approved):

apcd-cms/src/apps/view_users/views.py
```diff
@@ -36,93 +48,72 @@ def _edit_user(form):
...
             context['selected_status'] = status_filter
             queryStr += f'&status={status_filter}'
-            table_entries = table_filter(status_filter, table_entries, 'status', False)
+            # table_entries = table_filter(status_filter, table_entries, 'status', False)
 
-        if role_filter is not None and role_filter !='All':
+        if role_filter is not None and role_filter != 'All':
             context['selected_role'] = role_filter
-            table_entries = table_filter(role_filter, table_entries, 'role_name', False)        
+            # table_entries = table_filter(role_filter, table_entries, 'role_name', False)
```
rstijerina, line 105: Do we want to keep the commented lines of code in this file diff?

mjames-uth: I will remove those

mjames-uth: @rstijerina , comments removed

wesleyboar:
@mjames-uth @sophia-massie,

This PR is very old. Is it still relevant (given the conversion to React)?[^1]

[^1]: I admit I don't know the codebase well, just checking on old PRs (including my own).

PR #231, opened 2023-10-17
disable old reg form

rstijerina:
## Overview

Disables the `/register/request-to-submit/` path.


edmondsgarrett (approved):
Got a 404 for this url, LGTM! Thanks Sal

PR #232, opened 2023-10-17
Task/wp 284 submitter listing table

sophia-massie:
## Overview

The goal of this task is to create a new submitter_renewal_listing app to show a listing of all registrations for a submitter. The app:

- checks the role for submitter_admin
- grabs the submitter code
- grabs all distinct registration_ids filtered with the submitter code so only that submitter's organization's registrations are shown. 

## Related

- [WP-284](https://jira.tacc.utexas.edu/browse/WP-284)

## Changes
- In apcd_database.py:

1. Changed function name to get_submitter_info(user) so that multiple apps can use it and it be clear what the function does
2. Changed the get_registrations, get_registration_entities, and get_registration_contacts so that if a submitter_id is passed to the function, it will join the the submitter table with the registration and get all registration ids with a the matching submitter code.

- Created new submitter_renewals_listing app. Copied and pasted the admin_regis_table - the javascript was too fragile to put conditionals in the admin_regis_table view and template. 
**_- REASON:_** Since there is a tight deadline, Tracy requested that the admin_regis_table be completely copied and modified into its own app. Since we are planning to rework into React, she requested that the reuse of code be done when we switch to React rather than now.
- Changed wording  of header and description per Jodie at UTH
- Reused admin view modal and removed submitter view modal
- Made SubmittersTable a sub class of registrationsTable view
- Added drop down to link to form (This doesn't work yet until it's merged with Garrett's PR)

## Testing

1. Change APCD DB to prod credentials - test DB does not have year as a column on the registrations table yet. Joe is working on this
2. Go to http://localhost:8000/register/list-registration-requests/
3. Ensure the page looks like the screen shot. It doesn't look like TACCTEST submitter code has any associated entities or contacts. 
4. Double check that http://localhost:8000/administration/list-registration-requests/ is working as it previously was.

## UI

Screenshot 2023-10-25 at 12 40 34 PM


happycodemonkey (approved):
This looks good to me!

edmondsgarrett (approved):
New listing page looks good, everything is working well, thanks!

apcd-cms/src/apps/submitter_renewals_listing/templates/view_submitter_registration_modal.html
```diff
@@ -0,0 +1,115 @@
+<div id="viewRegistrationModal_{{r.reg_id}}" class="modal fade" role="dialog">
```
edmondsgarrett, line 1: This is a copy of `view-registration-modal.html`. I think we can instead use that template and load it here and for the admin list registrations table, for maintainability.

apcd-cms/src/apps/submitter_renewals_listing/static/submitter_renewals_listing/css/table.css
```diff
@@ -0,0 +1,14 @@
...
+/* SEE: https://css-tricks.com/responsive-data-tables/ */
+@media (max-width: 767px) {
+    /* To label the cells */
+    /* RFE: Add `data-label` to each cell so we can use `attr(data-label)` */
+    .submitters-table td:nth-of-type(1):before { content: "Business Name"; }
+    .submitters-table td:nth-of-type(2):before { content: "Type"; }
+    .submitters-table td:nth-of-type(3):before { content: "Location"; }
+    .submitters-table td:nth-of-type(4):before { content: "Registration Status"; }
+    .submitters-table td:nth-of-type(5):before { content: "Actions"; }
+}
```
edmondsgarrett, line 14: Can we reuse the admin registration table styles here as well?

apcd-cms/src/apps/submitter_renewals_listing/views.py
```diff
@@ -0,0 +1,229 @@
...
+from apps.utils.utils import table_filter
+from apps.components.paginator.paginator import paginator
+import logging
+import json
+from dateutil import parser
+
+logger = logging.getLogger(__name__)
+
+
+class SubmittersTable(TemplateView):
```
edmondsgarrett, line 15: Note for later: I think this class can be an instance of the registrations table class, and then super over the relevant methods (get, dispatch) and edit some of the context (e.g. the `pagination_url_namespaces`)

edmondsgarrett (changes requested):
Some changes for optimization below

apcd-cms/src/apps/submitter_renewals_listing/views.py
sophia-massie: Great point! Done.

apcd-cms/src/apps/submitter_renewals_listing/templates/list_submitter_registrations.html
```diff
@@ -0,0 +1,126 @@
...
+                  <td>{{r.biz_name}}</td>
+                  <td>{{r.type}}</td>
+                  <td>{{r.location}}</td>
+                  <td>{{r.reg_status}}</td>
+                  <td>
+                    {% include "view_registration_modal.html" %}
+                    <select id='actionsSubmitterDropdown_{{r.reg_id}}' class='status-filter' onchange="openAction('{{r.reg_id}}')">
+                      <option value="">Actions</option>
+                      <option value="viewRegistration">View Record</option>
+                      <option value="renewRegistration">Renew Registration (coming soon)</option>
```
edmondsgarrett, line 56: Should this be disabled at this point? Clicking this option points to a broken link.

edmondsgarrett (changes requested):
Page looks to be working, nice! Thanks for addressing those comments. One comment below:

apcd-cms/src/apps/submitter_renewals_listing/templates/list_submitter_registrations.html
mjames-uth: Agree with this.

mjames-uth:
I agreed on a comment a Garett left, otherwise, looks good

sophia-massie:
> I agreed on a comment a Garrett left, otherwise, looks good

Since the submitter form PR is merged, I didn’t disable the renew option because the script above should just work when it’s merged with your PR. Is it ok if I leave it enabled so there won’t have to be another PR to enable it @mjames-uth

apcd-cms/src/apps/submitter_renewals_listing/templates/list_submitter_registrations.html
```diff
@@ -0,0 +1,126 @@
...
+      selectedOption = actionsDropdown.options[actionsDropdown.selectedIndex].value;
+      modal_id = `${selectedOption}Modal_${reg_id}`;
+      $(`#${modal_id}`).modal({backdrop: "static"}); /* modal appears manually */
+      actionsDropdown.selectedIndex = 0; /* resets dropdown to display 'Select Action' again */
+      if (selectedOption == "renewRegistration") {
+        var xhr;
+        xhr = new XMLHttpRequest();
+        xhr.open('GET', `/register/request-to-submit/${reg_id}`)
+        xhr.send()
+        window.location.href = `/register/request-to-submit/${reg_id}`;
```
edmondsgarrett, line 120:
```suggestion
        xhr.open('GET', `/register/request-to-submit/?reg_id=${reg_id}`)
        xhr.send()
        window.location.href = `/register/request-to-submit/?reg_id=${reg_id}`;
```
I think the registration form view is looking for the var name `reg_id`, so this should fix that to where the link is no longer broken.

mjames-uth (approved):
Working as expected

mjames-uth (changes requested):
Can you add the registration_year column to the view of the List Registrations page. That will be needed to differentiate between the old registration and the the new registration.

edmondsgarrett (changes requested):
Listing is working and Renew action redirects to the form with pre-populated data, nice! I think we can then finish up this task in this PR with a couple more things (in addition to Mariah's requests):
- Can you remove the "(coming soon)" from the drop down action text for the Renew?
- Not sure if I have the correct submitter code for this, but I submitted a renewal from this action and said listing does not contain that new record. Is this expected?

sophia-massie:
New testing step

In secrets, use prod DB credentials. Testdb has not been updated to include a year column yet.

edmondsgarrett (approved):
LGTM, thanks for all the additions and changes

PR #235, opened 2023-10-23
bug/WP-340: Fix for add/remove entity buttons on edit registration

edmondsgarrett:
## Overview
Fixes bug where add/remove buttons for entities were not responding when editing a registration on Admin List Registrations page
…

## Related

- [WP-340](https://jira.tacc.utexas.edu/browse/WP-340)
…

## Changes
- Moved a html templating {% if %} statement outside of its previously-surrounding JSX string literal `${}` to resolve bug with compact if-else (`? :`)
…

## Testing

2. On any record, select 'Edit Record' from the Actions drop down
3. Try both adding and removing entities from the record with the corresponding buttons and confirm said buttons are responding
4. Bonus: try submitting the form with an additional/removed entity and confirm the submission is successful and, upon reload, the same record now reflects the submitted change

## UI

…

### DIFF
--- a/apcd-cms/src/apps/registrations/templates/submission_form/registration_form_scripts.html
+++ b/apcd-cms/src/apps/registrations/templates/submission_form/registration_form_scripts.html
@@ -345,12 +345,12 @@
           <div id="help-text-total_claims_value_${entities}" class="help-text">
           </div>
         </div>
-        ${ entities === 5 ?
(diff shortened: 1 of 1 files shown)


sophia-massie (approved):
LGTM!

mjames-uth (approved):
All good

PR #239, opened 2023-10-31
task/WP-347: Restrict registration renewal access

edmondsgarrett:
## Overview
With passing reg_id as a url parameter to the registration form to initiate a renewal, need to check both user role and reg_id validity for said user before a renewal can be initiated on the form.

## Related

- [WP-347](https://jira.tacc.utexas.edu/browse/WP-347)
…

## Changes
- If `reg_id` parameter is in GET request for the registration form, now also checks if a user if either a submitter admin or an APCD admin
- Then grabs user's submitter code and checks if queried `reg_id` is within submitter code's registrations before loading registration's data to the form
- If any of the above tests are not passed, instead loads registration form with no pre-loaded data
- Some code clean-up + linting found while working on main task 
…

## Testing

1. Go to http://localhost:8000/admin and:
    - Go to click users and your username
    - Remove submitter_admin from your groups and leave only APCD_ADMIN
    - Click save
    - Bring down your containers and bring them back up
2. Go to https://localhost:8000/register/request-to-submit/?reg_id=1 and confirm the page pre-loads with the registration data (no entities or contacts expected for this record)
3. Now try hitting https://localhost:8000/register/request-to-submit/?reg_id=5 and confirm the form loads, but with no data pre-loaded
4. Repeat step 1, but leave your role as only SUBMITTER_ADMIN and then repeat steps 2 and 3 and confirm the same
5. Repeat step 1, but leave your role as SUBMITTER_USER only and repeat steps 2 and 3; confirm that, in either case, the form does not pre-load with the data

## UI

…

### DIFF
--- a/apcd-cms/src/apps/registrations/views.py
+++ b/apcd-cms/src/apps/registrations/views.py
@@ -1,6 +1,7 @@
 from apps.utils import apcd_database
 from apps.utils.apcd_groups import has_apcd_group
 from apps.utils.registrations_data_formatting import _set_registration
+from apps.submitter_renewals_listing.views import get_submitter_code
 from django.conf import settings
 from django.http import HttpResponse, HttpResponseRedirect
 from django.template import loader
@@ -8,6 +9,7 @@ from django.views.generic import View
 from requests.auth import HTTPBasicAuth
 import logging
 import rt
+import json
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


apcd-cms/src/apps/registrations/views.py
```diff
@@ -21,13 +23,21 @@ class SubmissionFormView(View):
...
         renew = False
-        if 'reg_id' in request.GET:
-            reg_id = request.GET.get('reg_id')
-            renew = True
-            registration_content = apcd_database.get_registrations(reg_id)[0]
-            registration_entities = apcd_database.get_registration_entities(reg_id)
-            registration_contacts = apcd_database.get_registration_contacts(reg_id)
-            formatted_reg_data = _set_registration(registration_content, registration_entities, registration_contacts)
+        reg_id = request.GET.get('reg_id', None)
+        if reg_id and (apcd_database.get_user_role(request.user.username) in ['APCD_ADMIN', 'SUBMITTER_ADMIN']):
```
sophia-massie, line 27: I tried putting in a hardcoded real user name ('thbrown' for example) here and on line 29 in lieu of the request.user to test. When I clicked the renewal action in the listing dropdown, the url passed me on to a blank form template. Great work! 

On the form page, the url has a query param of reg_id=1 even though the form is blank. Will this form on post take this query param at any point? Maybe the reg_id parameter should be cleared after this check in the registration view if it doesn't pass the requirements for pre-populating?

apcd-cms/src/apps/submitter_renewals_listing/views.py
```diff
@@ -40,9 +39,10 @@ def get_context_data(self, registrations_content, registrations_entities, regist
...
         return context
-    
+
+
 def get_submitter_code(request):
     submitter = get_submitter_info(str(request))
     for i in submitter:
-        submitter_code =  i[1]
-    return JsonResponse(({'submitter_code' : submitter_code } if submitter_code else ""), safe=False)
+        submitter_code = i[1]
```
sophia-massie, line 47: 👍 nice catch

apcd-cms/src/apps/submitter_renewals_listing/views.py
```diff
@@ -40,9 +39,10 @@ def get_context_data(self, registrations_content, registrations_entities, regist
...
+    return JsonResponse(({'submitter_code' : submitter_code} if submitter_code else ""), safe=False)
```
sophia-massie, line 48: 👍 nice catch

happycodemonkey (approved):
Looks great!

apcd-cms/src/apps/registrations/views.py
edmondsgarrett: Good point, I'll look into it!

sophia-massie (approved):
LGTM!

PR #241, opened 2023-11-02
Task/wp 372 admin except extend entity name

sophia-massie:
## Overview

Updated Exception and Extension Admin tables so that it displays the extension and exception id's entity_name rather than org_name from the submitters table. 

I left the parameter for filtering by entity as org for brevity's sake in the URL. Let me know if it makes more sense to change to org. Or we can can check with Tracy and UTH.

## Related

- [WP-123](https://jira.tacc.utexas.edu/browse/WP-372)

## Changes

apcd_database.py:
- Changed get_all_exceptions and get_all_extensions queries to get entity name rather than org name

views.py:
- Changed the AdminExceptionTable and AdminExtensionTable class views so it sets the entity name from the DB response to "entity_name".
- Changed the AdminExceptionTable and AdminExtensionTable class views so it sets the header to Entity rather than Organization
- Appends entity_name from DB to org_options for the org_filter. Let these as org so the URL with query params were not as long
- Filter table by entity_name if org_filter is selected

Templates and Static:

- Changed styles so that when window is small, it changes columns to rows with Entity rather than Organization as the header
- Removed {{r.org_name}} and replaced with {{r.entity_name}}

## Testing

1. Go to [http://localhost:8000/administration/list-extensions/](http://localhost:8000/administration/list-extensions/)
2. Check that table matches UI below pulling entity_name rather than org_name
3. Check that filters still work as expected
4. Click view and edit actions to make sure entity_name is showing rather than org_name
5. Submit an edit and make sure it works
6. Go to [http://localhost:8000/administration/list-exceptions](http://localhost:8000/administration/list-exceptions)
7. Repeat steps 2-5

## UI
Exceptions Table
Screenshot 2023-11-02 at 3 00 00 PM

Extensions Table
Screenshot 2023-11-02 at 4 04 26 PM

…

### DIFF
--- a/apcd-cms/src/apps/admin_exception/static/admin_exception/css/table.css
+++ b/apcd-cms/src/apps/admin_exception/static/admin_exception/css/table.css
@@ -7,7 +7,7 @@
     /* To label the cells */
     /* RFE: Add `data-label` to each cell so we can use `attr(data-label)` */
     .exception-table td:nth-of-type(1):before { content: "Created"; }
-    .exception-table td:nth-of-type(2):before { content: "Organization"; }
+    .exception-table td:nth-of-type(2):before { content: "Entity"; }
     .exception-table td:nth-of-type(3):before { content: "Requestor Name"; }
     .exception-table td:nth-of-type(4):before { content: "Exception Type"; }
     .exception-table td:nth-of-type(5):before { content: "Outcome"; }
--- a/apcd-cms/src/apps/admin_exception/static/admin_exception/css/table.css
+++ b/apcd-cms/src/apps/admin_exception/static/admin_exception/css/table.css
@@ -1,16 +1,13 @@
-.modal-cell {
-    padding-left: 0px;
-}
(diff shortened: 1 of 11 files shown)


happycodemonkey:
Hey Sophia, I noticed "None" is showing up twice in the dropdown for Extension Outcome, unsure if that's a weird bug or something.

happycodemonkey (approved):
We are leaving the "None" drop down issue for further discussion and a new task, this change addresses the original task and looks great.

sophia-massie:
> We are leaving the "None" drop down issue for further discussion and a new task, this change addresses the original task and looks great.

Should I remove the last changes from commit #[73902dc](https://github.com/TACC/Core-CMS-Custom/pull/241/commits/73902dc2c43a9ee2af70867fc07d2e8ed9ea0528) @happycodemonkey ?

edmondsgarrett (approved):
LGTM!

PR #242, opened 2023-11-06
hotfix/wp-375-extension-entity-name-add-button

sophia-massie:
## Overview

Entity name missing from extension form add buttons

## Related

- [WP-375](https://jira.tacc.utexas.edu/browse/WP-375)

## Changes
Added entity_name var into template rather than org_name

## Testing

1. Go to [http://localhost:8000/submissions/extension-request/](http://localhost:8000/submissions/extension-request/)
2. Click Add Extension button
3. Make sure drop down populates with an entity name

### DIFF
--- a/apcd-cms/src/apps/extension/templates/extension_submission_form/extension_submission_error.html
+++ b/apcd-cms/src/apps/extension/templates/extension_submission_form/extension_submission_error.html
@@ -12,7 +12,7 @@
     <p class="c-message c-message--error">
         An error occurred during your extension request submission. For help, <a href="/workbench/dashboard">submit a ticket</a>.
     </p>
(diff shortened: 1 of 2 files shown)


mjames-uth (commented):
Found 1 error: In the event of an error: theres a "Go Back to Extension" button. It did not work. When clicked there is no action

sophia-massie:
> Found 1 error: In the event of an error: theres a "Go Back to Extension" button. It did not work. When clicked there is no action

Just pushed up a new changed to prevent this. Thanks!

happycodemonkey (approved):
Looks great

mjames-uth (approved):

PR #246, opened 2023-11-08
task/update-sub-admin-query-table

sophia-massie:
## Overview

To use the new registration_submitters table to allow a submitter code to have a many to many relationship with registration ids on the registration table

## Changes

Joined the registration_submitters table to link multiple registration ids to the submitter table

## Testing

1. Unable to test locally until UTH starts using the registration_submitters table. The query will pull all distinct registration ids for a submitter_code once they are on the registration_submitters table.

## UI

…


happycodemonkey (approved):
We added some data to test locally and it looks great to me

edmondsgarrett (approved):
Looks good! Tested locally (thank you Carrie)

PR #247, opened 2023-11-08
task/WP-345--entities-changes-uth---proposed-query-fix

sophia-massie:
## Overview

Adjust entities and contacts queries so view displays appropriate contacts and entities for a registration id when passed to the prepopulated form, listing view modal, and submitter listing table 

## Related

- [WP-345](https://jira.tacc.utexas.edu/browse/WP-345)

## Changes

- For each registration, pass in the reg_id to get the entities and contacts associated to that registration
- Then, pass in submitter_code to get all registrations for a submitter_admin submitter_code to display on the submitter admin page

## Testing
_Most of this review will involve looking at the code and database if you have access to ensure the pages and queries are pulling the correct things._

1. Branch off [task/WP-345--entities-changes-uth  #245](https://github.com/TACC/Core-CMS-Custom/tree/task/WP-345--entities-changes-uth) and checkout this branch 
**_If #245 has been merged into main already, skip this step_**
2. Go to [http://localhost:8000/register/list-registration-requests/](http://localhost:8000/register/list-registration-requests/)
3. Check that entities and contacts are showing up in the view modal
4. Check that the same entities and contacts show up when you click on the renew action in the drop down.
5. The form should be pre-populated for each entity except for the Coverage Estimates section

### DIFF
--- a/apcd-cms/src/apps/submitter_renewals_listing/views.py
+++ b/apcd-cms/src/apps/submitter_renewals_listing/views.py
@@ -18,6 +18,12 @@ class SubmittersTable(RegistrationsTable):
             data = json.loads(submitter_code)
             submitter_code = data['submitter_code']
             registrations_content = get_registrations(submitter_code=submitter_code)
+            registrations_entities = []
+            registrations_contacts = []
+            for reg in registrations_content:
+                reg_id = int(reg[0])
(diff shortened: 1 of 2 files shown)


edmondsgarrett (approved):
Already tested between us, works well!

PR #268, opened 2024-02-09
Bug/WP-483 - APCD fixes for extensions

chandra-tacc:
## Overview

Submitter calendar can accept duplicates by db design, but non functional in use cases. Business logic / queries need to adjust to that.
The key fields in submitter_calendar are submitter_id, data_period_start. 
Submitter_id is distinct for each submitter (org, payor_code). There is no other information in submitter_calendar to derive uniqueness. With the current db design, it is safe to assume the data_period_start should be unique per submitter_id.
If any of the business logic needs to change, that will require db design change.

## Related
[WP-483](https://tacc-main.atlassian.net/browse/WP-483)

## Changes

* For time period use distinct
* Change button text to go to dashboard

## Testing

1. Uploaded duplicate rows in test DB:

`select calendar_item_id, submitter_id, data_period_start, expected_submission_date from submitter_calendar;`

Screenshot 2024-02-08 at 5 35 22 PM

Screenshot with no duplicates in UI.   
Screenshot 2024-02-08 at 5 30 08 PM

2. Tested the exception and successful one shows 'go to dashboard'

Screenshot 2024-02-08 at 5 30 48 PM

Also, navigation took to the dashboard

### DIFF
--- a/apcd-cms/src/apps/extension/templates/extension_submission_form/extension_form_success.html
+++ b/apcd-cms/src/apps/extension/templates/extension_submission_form/extension_form_success.html
@@ -8,6 +8,6 @@
     <p>
         Your extension submission was successful.
     </p>
-    <a class="c-button c-button--primary" href="/submissions/extension-request/">Go Back to Extensions</a>
+    <a class="c-button c-button--primary" href="/workbench/dashboard">Go to Dashboard</a>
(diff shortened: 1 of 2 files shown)


sophia-massie (approved):
LGTM thank you

PR #269, opened 2024-02-09
First version of OpenApi spec based ApiClient and integration into model

chandra-tacc:
## Overview
This PR attempts to use APCD api spec to perform CRUD operations on APCD objects.

APCD api openapi spec is available. There are some strategies:

* Dynamically Loading OpenAPI Specs:
   * No generated code in source control
   * Easily handle updates, by generating this during runtime. Example: tapipy does this today.
   * generating dynamically would be involve:
     * using openapi lib to read spec.
     * setattr to object for each operation. Example client.submission.get (would be submission attr in client and then get/post attrs in submission)
     *  And write a wrapper to send request for each operation. Example from [tapipy](https://github.com/tapis-project/tapipy/blob/main/tapipy/tapis.py#L968). This code handles missing required inputs.

  * Cons: perf is slower, because this is generated runtime.
  * Cons: no static typing
* Statically generating stubs using OpenAPI Specs:
  * Static Typing
  * better performance.
  * Cons: large set of generated code. A strategy to overcome this is to generate this during build time and do not submit the code to source control.
## Related

[WP-476](https://tacc-main.atlassian.net/browse/WP-476)
## Changes

Using compile time generation

1) Generate client using openapi spec generator
2) With token refresh, use the client to get new token
3) provide a method to get an authenticated client

## Testing

* Did not test yet. Note: this requires toml changes to include pydantic which openapi uses.


rstijerina (approved):
Approving, though I believe a packaged solution is also being considered

PR #275, opened 2024-02-15
WP-494: handle null entity_names when using exception and extensions in admin view

chandra-tacc:
## Overview
There is bad data in db with empty entity name in submitters table. This is leading to 500 errors in UI.

## Related

- [WP-494](https://tacc-main.atlassian.net/browse/WP-494)

## Changes
* Fix DB queries so any business logic using the data is not broken - check for both null and empty
* Do not allow empty strings in dropdown.

## Testing

* Entity name empty strings show up in UI table listing
* Drop down listing filter out empty entity name.

1. Admin exception listing - the listing happens with empty entity_name
Screenshot 2024-02-15 at 12 15 25 PM

2. Admin extension listing - the listing happens with empty entity_name
Screenshot 2024-02-15 at 12 15 11 PM


edmondsgarrett (approved):
LGTM! Thanks Chandra

happycodemonkey (approved):
Looks good!

PR #276, opened 2024-02-20
Initial commit - using react with APCD admin registration list page

chandra-tacc:
## Overview

The goal is to use react components in APCD custom pages.

The react setup is as follows:
* A client that generates a lib of APCD components.
* Requests to backend are made through fetchUtil (similar to core portal client)
* The entry point provides setup function to render the component
* The client includes TUP components.
* Docker uses build_client script to build and copy artifact (closely following core portal build process)
* The template file includes the lib file and sets up the relevant component.

### Steps for each APCD page
* Create a component in client
* Include the component in library.tsx and add a setup method to render the component in library.tsx
* Add react-root in template html
* Adjust the template html to import and render the component

* Looking for feedback or ideas on the following: 
  * Whether this multi-page react app approach is accurate. 
  * Client structure - lib with component setup method.
  * vite usage.

## Related

[WP-476](https://tacc-main.atlassian.net/browse/WP-123)

## Changes

* Change the django view to serialize the content in json
* Create top level client folder.
* Add package.json, vite and tsconfig related file. Note: Vite hmr usage is not setup yet.
*  Add tup-components, hooks, components, utils in client
* Copy TUP-components from [TUP Repo](https://github.com/TACC/tup-ui/tree/main/libs/core-components/src )
* Add components/Admin/Registrations/AdminRegistrations.tsx for Admin Table
* Add hooks/admin/useAdmin.ts for admin registration listing.
* Add util/fetchUtil.js to handle server requests.
* Add library.tsx to provide entry point for using in template.

### Missing

* unit test setup is not done.
* Cache busting and vite local dev setup is not done.
* whether to generate umd file or not
* Admin registration table feature is not complete. This is just first page.
* Any optimizations on react imports in django.

## Testing

Visit http://localhost:8000/administration/list-registration-requests/

Screenshot 2024-02-20 at 1 05 18 PM

network request:
Screenshot 2024-02-20 at 1 05 31 PM

## UI
Initial commit, trying not to modify UI. If styles are not matching, still working to get it to match.


apcd-cms/src/apps/admin_regis_table/templates/list_registrations.html
```diff
@@ -33,35 +40,7 @@ <h1>List Registrations</h1>
...
-                      <option value="">Select Action</option>
-                      <option value="viewRegistration">View Record</option>
-                      <option value="editRegistration">Edit Record</option>
-                    </select>
-                  </td>
-              </tr>
-          {% endfor %}
-      </tbody>
-  </table>
+  <div id="react-root"></div>
```
wesleyboar, line 43: Nice. Rather than implement Portal and mimic URLs, load the React app onto the existing pages. 👍

wesleyboar (commented):
I request `shared-components` be `common-components`, like other repo instances[^1], because [TUP-700](https://tacc-main.atlassian.net/browse/TUP-700) will create `@tacc/core-components` Node package to replace the copy-pasting.

[^1]: https://github.com/TACC-Cloud/hazmapper/tree/master/react/src/core-components and https://github.com/TACC/tup-ui/tree/main/libs/core-components and (similar to) https://github.com/TACC/Core-Portal/tree/main/client/src/components/_common

chandra-tacc:
> Should this use Vite, like hazmapper React and tup-ui? @jarosenb is the progenitor of it at TACC, I think.

It is using vite (see[ package.json](https://github.com/TACC/Core-CMS-Custom/pull/276/files#diff-36cca132e7a8b2dd4b9080f512153ae0b4d4d050c29b87727a5748f12baf00ae)). 
I did not set it up for hmr yet. Will work on it once we all agree on this approach.

wesleyboar:
Vite. Whoops, missed it. Cool.

Also, me typo: `shared-components` → <del>`common-components`</del> <ins>`core-components`</ins>

edmondsgarrett (approved):
Looks good, I think this structure will work well for this project. Thanks for this work!

chandra-tacc:
> Vite. Whoops, I missed it. Cool.
> 
> Also, me typo: `shared-components` → ~`common-components`~ `core-components`

Done renamed

wesleyboar (commented):
Thanks for d870dec rename. ☺ The code and screenshot suggests UI is not changed. ✓ [My test fails](https://tacc-team.slack.com/archives/C03B7899YD8/p1708976849285909) — state switches form loading to error. (Maybe it is a local dev env issue, like settings.)

| State 1 | State 2 |
| - | - |
| error | loading |

apcd-cms/src/client/src/components/library.tsx
sophia-massie, line 1: Not a blocker - just an idea that can be follow on work if you like it.

Would it make sense to create a query client wrapper in a follow on PR so that each component will not have to have this repeated code? Then have a higher level main.tsx file that implements it?
See in TUP:
https://github.com/TACC/tup-ui/tree/main/libs/core-wrappers/src/lib/QueryWrapper
https://github.com/TACC/tup-ui/blob/main/apps/tup-ui/src/main.tsx

chandra-tacc: @sophia-massie  - that is a good idea. I did not spend time looking into routing (the main.tsx -> app.tsx from link above details the routing), we could try that and see what happens when we do not have a default mainly because this Multi Page app vs SPA. 
Regarding QueryWrapper, yes, good idea to use it.
Created task https://tacc-main.atlassian.net/browse/WP-506 to track it.

sophia-massie (approved):
Thank you so much for this proof of concept, Chandra. LGTM 👍

apcd-cms/src/client/src/components/library.tsx
sophia-massie: Oh right the routing may or may not be a good fit for this project but we can look in follow on work.

PR #281, opened 2024-03-27
WP-519: APCD Custom getting started page

chandra-tacc:
## Overview

Getting started for APCD needs to be same link for all 3 environments. Since this page is being modified for other areas, just add a custom template.

## Related

- [WP-519](https://tacc-main.atlassian.net/browse/WP-519)

## Changes

1) add a new template
2) deployment config is modified: https://github.com/TACC/Core-Portal-Deployments/blob/main/apcd/camino/cms.settings_custom.py#L23 

## Testing

1. Go to https://dev.apcd.tacc.utexas.edu/help/getting-started/
2. Click the signup for a new account link
3. It should be pointing to non-dev url


wesleyboar (commented):
If you put the template instead in `/apcd-cms/src/taccsite_cms/templates/guides/`, then there is **no need** to [update the settings](https://github.com/TACC/Core-Portal-Deployments/commit/654a5a3) **and also** the template path will remain standard and compatible. See #282 for change and more explanation.

chandra-tacc:
> If you put the template instead in `/apcd-cms/src/taccsite_cms/templates/guides/`, then there is **no need** to [update the settings](https://github.com/TACC/Core-Portal-Deployments/commit/654a5a3) **and also** the template path will remain standard and compatible. See #282 for change and more explanation.

Thanks @wesleyboar  - I took your changes and tested it on dev.apcd. It works.

wesleyboar (approved):

PR #282, opened 2024-03-27
WP-519: APCD Custom getting started page - more standard & compatible

wesleyboar:
## Overview

Overwrite template, instead of creating a new one, so that [settings need **not** be updated][new setting].*

<details><summary>What is wrong with updating the setting?</summary><br />

[Such a settings update][new setting] in my experience has **not** been backwards-compatible (prod page will break upon deploy) **nor** future-compatible (not using a custom template in future, prod page will break upon deploy).

### Fail Case

0. Deploy project
    - with guide template setting path "B"
    - to a server that guide template setting path "A".
1. Open existing page that uses "Guide: Getting Started" template.
2. Notice page _crash_…
    <sup>(**unless** someone updated the page after deploy via `/admin` UI to use guide template setting path "A" before actually visiting the page)</sup>

    fail case example

</details>

[new setting]: https://github.com/TACC/Core-Portal-Deployments/commit/654a5a3b
[old setting]: https://github.com/TACC/Core-Portal-Deployments/blob/7b86c8a/apcd/camino/cms.settings_custom.py#L23
[old settings]: https://github.com/TACC/Core-Portal-Deployments/blob/7b86c8a/apcd/camino/cms.settings_custom.py

## Related

- [WP-519](https://tacc-main.atlassian.net/browse/WP-519)
- simplifies #281

## Changes

- **moved** a template
    <sup>**added** text to distinguish template from the Core-CMS one it overwrites</sup>

### Local

1. Follow ["Start Project" instructions](https://github.com/TACC/Core-CMS-Custom/blob/00de03e/README.md#start-project).
2. Create/Update `/apcd-cms/src/taccsite_cms/settings_custom.py` with [pre–WP-519 settings][old settings].
3. Create a page.
4. Set template to "Guide: Getting Started".
5. Verify page content is from modified template.

### Remote

1. Deploy this to a server.
2. **If page is set to use new template, then** expect page to fail to load.
3. **If so, then** via `/admin` change template to original template path.
4. Notice new template is used (at old template path).

## UI

new content


chandra-tacc:
Thanks for this fix, works great. Will follow this in future.

PR #294, opened 2024-07-25
[ACPD React Conversion]: Setup vite local and add base template with react

chandra-tacc:
## Overview
Currently, every template needs to add setup code for react. This PR will add the following:
* Simpler setup for every page. The base template and main.tsx handle the setup.
* Local development setup with vite.

## Related
[WP-429](https://tacc-main.atlassian.net/browse/WP-429)

## Testing

for local development:
* make build
* make start
* cd client
* npm run dev

1. Go to http://localhost:8000/administration/list-registration-requests/ and see if it works
2. Go to other admin pages like view-users and ensure they work.

## UI

…


van-go (approved):
LGTM

PR #295, opened 2024-07-29
Adapt Makefile for docker v2

chandra-tacc:
## Overview
Docker v2 has different commandline for compose `docker compose`
Adjust Makefile to handle both based on availability
## Related

[WP-602](https://tacc-main.atlassian.net/browse/WP-602)

## Changes
Add command var based on whether `docker-compose` works

## Testing

1. Update to v2.
2. cd to apcd_cms
3. make build
4. make start

PR #301, opened 2024-08-07
fix: WP-602 `docker compose` with and without dash

wesleyboar:
## Overview

Docker v2 has different command for `docker compose`. Adjust Makefile to handle both based on availability.

## Related

- [WP-602](https://tacc-main.atlassian.net/browse/WP-602)
- mimics https://github.com/TACC/Core-CMS-Custom/pull/295

## Changes

- **added** conditional `docker compose` command to Makefile

## Testing

1. `make build` — no error firing command from Makefile
2. `make start` — no error firing command from Makefile

## UI

Skipped.


wesleyboar:
Merged based on it mirroring https://github.com/TACC/Core-CMS-Custom/pull/295 **identically** for three other projects.

PR #308, opened 2024-09-06
enhance: newest unofficial new lccf logo

wesleyboar:
## Overview

Update LCCF logo. Not the first time this year. Not the last time.

_Also, remove old placeholder logos, because lead designer is back._


wesleyboar:
Merging, because the code was deployed to prod.

PR #309, opened 2024-09-06
Hotfix: null handling in the status filter options for all admin tables

sophia-massie:
## Overview

Checks for null for filter options for listings in admin exceptions and admin extensions.

## Related
[WA-349](https://tacc-main.atlassian.net/browse/WA-349)

## Changes

…

## Testing

1. Then `git checkout hotfix/check-for-null-status-extension`
2. Test http://localhost:8000/administration/list-exceptions/
3. http://localhost:8000/administration/list-extensions/

## UI

…


chandra-tacc (approved):
LGTM. Tested, ensured nulls are in dev db and also checked 'All' shows as the first entry.

edmondsgarrett (changes requested):
This PR solves the NoneType-str comparison bug, but the result is that the 'None' shows up in the filter options as the first element in the dropdown, so the page shows that 'None' is the selected status filter option even though 'All' records are shown.

PR #312, opened 2024-09-17
task/WP-695: Add status change to edit registration modal

edmondsgarrett:
## Overview
…

## Related

- [WP-695](https://tacc-main.atlassian.net/browse/WP-695)
…

## Changes
- Adds new `<select>` element to registration form, only rendered if registration data is being loaded to the form + is not a registration renewal
- Adds write to `registration_status` field on registrations table on database for `update_registration` db util function 
…

## Testing

2. Select a record to edit via Actions column -> "Edit Record"
3. Confirm that a new drop down labeled 'Registration Status' appears at the top of the 'Organization' portion of the form
4. Confirm this drop down's pre-selected value matches that in the 'Registration Status' column for the same record on the table
5. Select a different option in the drop down and then submit that change
6. Confirm you are directed to a success page, then return to the List Registrations page and confirm the value you selected shows in both the table for the record you edited + the same 'Edit Record' modal
7. Confirm this new field does not appear on the following pages:
   - http://localhost:8000/register/request-to-submit/ (reg form)
   - http://localhost:8000/register/request-to-submit/?reg_id=2 (reg renewal form)

## UI
image

…

### DIFF
--- a/apcd_cms/src/apps/registrations/templates/submission_form/registration_form_body.html
+++ b/apcd_cms/src/apps/registrations/templates/submission_form/registration_form_body.html
@@ -5,17 +5,18 @@
         <input type="hidden" name="edit-registration-form" value>
         <input type="hidden" name="reg_id" value="{{r.reg_id}}">
 
-        <label for="reg_status"> Registration Status </label>
-
-        <div class="field-wrapper select">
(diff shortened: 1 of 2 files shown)


chandra-tacc (approved):
tested and LGTM

sophia-massie (approved):
LGTM

PR #317, opened 2024-09-20
bug/WP-697-User-Org-Filter-Bug

sophia-massie:
## Overview

When None organization filter is selected on Admin View Users page, make sure None stays selected on load of the filter

## Related

- [WP-697](https://tacc-main.atlassian.net/browse/WP-697)

## Changes

Adjusts handling of query response to add string None to the organization options

## Testing

1. Go to http://localhost:8000/administration/view-users/?status=All&org=None and make sure None is displayed in the Organization drop down


chandra-tacc (approved):
Tested and LGTM

PR #319, opened 2024-09-20
hotfix/allow-zero-in-reg-forms

sophia-massie:
## Overview

Allow registration to allow zero in some numeric input fields

## Related

…

## Changes

Changed pattern in registration form so that the following allows 0 as an input

- total_covered_lives
- claims_encounters_volume
- total_claims_value

## Testing

1. Go to [http://localhost:8000/register/request-to-submit/](http://localhost:8000/register/request-to-submit/)
2. Input zero for the following fields in 2 different entity forms
    - total_covered_lives
    - claims_encounters_volume
    - total_claims_value
3. Go to [http://localhost:8000/administration/list-registration-requests/](http://localhost:8000/administration/list-registration-requests/) and select a test entry to edit
4. Input zero for the following fields in two different entities
    - total_covered_lives
    - claims_encounters_volume
    - total_claims_value
## UI

…

### DIFF
--- a/apcd_cms/src/apps/registrations/templates/submission_form/registration_form_body.html
+++ b/apcd_cms/src/apps/registrations/templates/submission_form/registration_form_body.html
@@ -463,7 +463,7 @@
                         class="integerfield"
                         id="total_covered_lives_{{ent_no}}_{{r.reg_id}}"
                         inputmode="numeric"
-             ^(0|[1-9][0-9]*)$
+                        pattern="^(0|[1-9][0-9]*)$"
                         value="{% if not renew %}{{ entity.no_covered }}{% endif %}"
                     />
                     </div>
@@ -482,7 +482,7 @@
                         class="integerfield"
                         id="claims_encounters_volume_{{ent_no}}_{{r.reg_id}}"
                         inputmode="numeric"
-             ^(0|[1-9][0-9]*)$
+                        pattern="^(0|[1-9][0-9]*)$"
                         value="{% if not renew %}{{ entity.claim_and_enc_vol }}{% endif %}"
                     />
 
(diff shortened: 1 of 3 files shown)


apcd_cms/src/apps/registrations/templates/submission_form/registration_form_body.html
```diff
@@ -463,7 +463,7 @@ <h6>
                         class="integerfield"
                         id="total_covered_lives_{{ent_no}}_{{r.reg_id}}"
                         inputmode="numeric"
-                        pattern="^[1-9]+[0-9]*$"
+                        pattern="^[0-9]*$"
```
chandra-tacc, line 466: can you use this instead : `^(0|[1-9][0-9]*)$`

`"^[0-9]*$"` allows a string like 0000. Instead we only want a string which is valid number.

apcd_cms/src/apps/registrations/templates/submission_form/registration_form_body.html
```diff
@@ -506,7 +506,7 @@ <h6>
                         class="integerfield"
                         id="total_claims_value_{{ent_no}}_{{r.reg_id}}"
                         inputmode="numeric"
-                        pattern="^(?=.*[1-9])[0-9]*[.]?[0-9]{1,2}$"
+                        pattern="^(0|[1-9][0-9]*)$"
```
chandra-tacc, line 509: The diff here is wrong, because the original pattern is lost. I think we need to do this one pattern at at time and allow 0|

chandra-tacc (approved):
LGTM.
Tested with 0, 0.00, 1.01, 0.01, etc.

PR #331, opened 2024-10-16
Bug/WP-711: Do not use hardcoded org names for user list

chandra-tacc:
## Overview
The list of org names is derived from actual data.

## Related
- [WP-711](https://tacc-main.atlassian.net/browse/WP-711)

## Changes

* Adjust the `view-users/api/options`  to read from DB.
* Since option api is available, remove the hard coded list from actual data.

## Testing

1.http://localhost:8000/administration/view-users/?org=None


van-go (approved):
LGTM

PR #333, opened 2024-10-20
Extension Form: Split into two components

chandra-tacc:
## Overview
Split Extension Form into two components - one parent and another child.
Formik has the context of the values in Field component, so no need for passing in values to child components.

## Testing

1. Go to http://localhost:8000/submissions/extension-request/
2. Enter data and submit, see console log


sophia-massie (approved):
LGTM and super helpful for my exception form! To get dynamic fields and styles, can use the Exception form for a follow on

PR #345, opened 2024-11-04
Task/wp 743 exception form testing changes

sophia-massie:
## Overview

Addresses bugs and UI fixes from APCD testing session 

## Related

- [WP-743](https://tacc-main.atlassian.net/browse/WP-743)

## Changes

…

## Testing

1. Go to [http://localhost:8000/submissions/exception/](http://localhost:8000/submissions/exception/)
2. Make sure that the UI has been updated to reflect these issues ( see Jira issue for another view of the breakdown of changes)

- [ ] If you change from a selected file type back to “-- Choose File Type ---”, it does not clear the field code drop down from cdls from previous selection. It resets to – Select Field Code --, but still has previous options from first selection

- [ ] Should notify if there are form field errors and it does not submit next to submit button.

- [ ] change email validation error to match extension form

- [ ] Change select business name default to match extension form

- [ ] On both types, it should not box the phrase “Your changes will not be saved if you change the exception type.” That box look just like the pull down box to change type. Use text only with a bold word “Note: Your changes will not be saved if you change the exception type.” 

- [ ] Mark fields as required with core-components Badge like CEP

- [ ] After selecting a business and then selecting “Select a Business” again the following error is shown:
- exceptions[0].businessName must be a `number` type, but the final value was: `NaN` (cast from the value `"-- Select a Business --"`).

- Address confusing requested threshold validation message when zero. : Requested Threshold Percentage - This message is confusing: “Must be less than the 0 required.” I tried 0.25, .25, 25, and 0. None worked. I wasn’t able to submit the form. 

#### Threshold page

Screenshot 2024-11-04 at 5 11 38 PM

### Other exception page

Screenshot 2024-11-04 at 5 14 07 PM

### DIFF
--- a/apcd-cms/src/apps/exception/templates/exception_submission_form.html
+++ b/apcd-cms/src/apps/exception/templates/exception_submission_form.html
@@ -3,7 +3,6 @@
 
 {% block content %}
 
-<link rel="stylesheet" href="{% static 'forms/css/exception_submission_form.css' %}">
 <div class="container">
   {% include "nav_cms_breadcrumbs.html" %}
       <div id="exception-submission-root">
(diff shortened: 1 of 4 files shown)


chandra-tacc (approved):
Tested all the steps suggested.

PR #346, opened 2024-11-05
Bug/WP-749: ViewUsers - fix filter usage.

chandra-tacc:
## Overview

* Filter select fields should be max width of the contents
* Filtering are not appending query params to url.
* Filters need spacing between them - alignment is proper.
* Header 'view more' should be 'actions'.

## Related
[WP-749](https://tacc-main.atlassian.net/browse/WP-749)

## Changes
1. adjustment to how data is fetched, so the status and org selections are handled correctly
2. clear selection works correctly.

## Testing

1. Go to http://localhost:8000/administration/view-users/
   * Test that the page loads
   * filters work
   * clear filter works.

## UI
Screenshot 2024-11-05 at 10 59 42 AM


happycodemonkey (approved):
Looks great!

PR #365, opened 2024-11-18
Pending for default status

happycodemonkey:
## Overview

Pending is now default for status. Also fixed the "Clear Options" button so it is consistent with other pages. 

## Changes

…

## Testing

1.

## UI

…


edmondsgarrett (approved):
LGTM, works as expected!

PR #368, opened 2024-11-20
WP-201: add entity organization to user file submissions

chandra-tacc:
## Overview
* Submitters can be attached to multiple organizations, so need to add this column to reflect this on user listing page for viewing submissions. 

## Related

[WP-201](https://tacc-main.atlassian.net/browse/WP-201)
## Changes

1. Fix query for user submission listing.
2. adjust response.
3. adjust table structure
4. fix the table class name 
5. fix the css
6. With entity organization in both submission listing, there is duplication of types and modal. Remove the duplicate modal, used single type system. Also, the admin submitter listing was missing proper css classes, fixed that.

## Testing

1. Go to http://localhost:8000/submissions/list-submissions/ and check the new column
2. reduce the size and check the column names.

## UI

Screenshot 2024-11-20 at 9 22 35 AM

Collapse view:
Screenshot 2024-11-20 at 9 22 25 AM


edmondsgarrett (approved):
Looks great! Thanks for the cleanup + removing the duplicate component.

PR #369, opened 2024-11-20
task/WP-700:  Add status to view record modal + default filter registration records to 'Received' status

edmondsgarrett:
## Overview
…

## Related

- [WP-700](https://tacc-main.atlassian.net/browse/WP-123)

## Changes
- For data passed to component `ViewRecordModal`, added registration status to data object
- Set an initial status filter in `RegistrationList` component dependent on `isAdmin` prop:
   - 'Received' if True
   - '' if False

## Testing

1. Go to http://localhost:8000/administration/list-registration-requests/ and confirm 'Filter by Status' dropdown shows 'Received' and that all records on listing have 'Registration Status' as 'Received'
2. Go to 'Actions' -> 'View Record' and confirm 'Registration Status' is shown under 'Organization', below 'Mailing Address' (see UI image below)
3. Change status filter to 'All' and confirm listing reloads, then shows all records (verify by status column), and that the 'Clear Options' button appears next to filters
4. Repeat step 2
5. Go to http://localhost:8000/register/list-registration-requests/ and confirm step 2 + that status filter defaults to 'All' on load

## UI
image


happycodemonkey:
Extremely small comment, the status on the http://localhost:8000/register/list-registration-requests/ view modal is lower case (presumably showing the db value rather than the display value)

happycodemonkey (approved):
Looks great!

chandra-tacc (approved):
LGTM

PR #373, opened 2024-11-20
Task/WP-387: Fix path URLs to not contain regex

jmcmillenmusic:
## Overview

Docker gives warnings for us having regular expressions (regexes) in our URLs. This change prevents those warnings.

## Related

[WP-387](https://tacc-main.atlassian.net/browse/WP-387)

## Changes

I imported `re_path` and applied it to any `path` that contained a regex.

## Testing

1. Turn on the Docker containers using `make start` and fire up the client with `npm run dev` as usual.
2. Login to the admin site.
3. Navigate to any page.
4. Watch the Docker logs to see that there are no more warnings.

## UI

image

### DIFF
--- a/apcd_cms/src/apps/admin_regis_table/urls.py
+++ b/apcd_cms/src/apps/admin_regis_table/urls.py
@@ -1,10 +1,10 @@
-from django.urls import path
+from django.urls import path, re_path
 from apps.admin_regis_table.views import RegistrationsTable
 
 app_name = 'admin_regis_table'
 urlpatterns = [
     path('list-registration-requests/', RegistrationsTable.as_view(), name='admin_regis_table'),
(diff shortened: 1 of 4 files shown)


rstijerina (changes requested):
We actually want to use `path`, and change the urls to not use regex. Instead, they can use standard django path parameters: https://docs.djangoproject.com/en/4.2/topics/http/urls/

jmcmillenmusic:
> We actually want to use `path`, and change the urls to not use regex. Instead, they can use standard django path parameters: https://docs.djangoproject.com/en/4.2/topics/http/urls/

Gotcha. I figured both options from the JIRA ticket were on the table.
I'll go back and fix that.

PR #374, opened 2024-11-21
WP-754: submissions - html and json log download

chandra-tacc:
## Overview
1. Allow user to download html and json logs for submissions.
2. Must work for both admin and submitter user

## Related

[WP-754](https://tacc-main.atlassian.net/browse/WP-754)

## Changes

1. Adjust db query to let submissons list know whether the log file is available or not for both html and json
2. Write a method to download log file for admin and submitter user
3. write view to process request and provide log with content disposition
4. hook to urls and get method in both admin and submitter
5. adjust ui to allow for download.

## Testing

When log is downloaded, the file name of the download should match the name in db.
1. submitter listing and download log: http://localhost:8000/submissions/list-submissions/
2. admin listing and download log: http://localhost:8000/administration/list-submissions/
3. If log is not available, do not show download link
4. Going to some other users log_id in browser should fail. http://localhost:8000/submissions/view_log?log_type=html&log_id=305

## UI

Screenshot 2024-11-21 at 10 32 32 AM


sophia-massie (approved):
Tested and LGTM!

PR #379, opened 2024-11-25
Task/WP-480: Submitter User List Registrations

jalowe13:
## Overview
A new page - Requested Registrations - by user, not Submitter/Org. We maybe put a note on this page, this list is only UNassigned organization.
…

## Related

…

## Changes

…

## Testing

1.

## UI

…

### DIFF
--- a/apcd_cms/src/apps/test_page/apps.py
+++ b/apcd_cms/src/apps/test_page/apps.py
@@ -1,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class TestPageConfig(AppConfig):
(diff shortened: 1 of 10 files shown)


jalowe13:
Should users see their registration requests immediately after filling out the form or after they are processed?

jalowe13:
After further investigation, users are not connected to their registrations in the database until they are linked as a submitter. A new field would need to be made in the database to reflect this. A new field on the users table that stores registration id’s would be a potential solution. This needs to be requested from UTHealth and checking if the process for creating submitters from a registration has been finalized.

PR #382, opened 2024-12-04
Bugs/WP-789: Extension Edit Modal

chandra-tacc:
## Overview
Fixes the following issues:
1. Remove references to `Exception` and `Users` in the labels and code.
2. After error, clear it when re-submitted
3. Disable submit when nothing is edited
4. Date periods are previously retrieved based on what the user has access to, instead get it for the submitter_id - this is admin page, so it should retrieve all available.

## Related

[WP-789](https://tacc-main.atlassian.net/browse/WP-789)

## Changes
1. One significant change is to add new api to retrieve available date period for submitter id.

## Testing

1. Go to http://localhost:8000/administration/list-extensions
2. Use the edit modal and try the following
     * labels are accurate
     * submit with bad date and error should go away on resubmit
     * submit is disabled if there is no edit
     * date periods for extensions that current user did not submit should still retrieve date periods and populate in UI.

### DIFF
--- a/apcd_cms/src/apps/admin_extension/views.py
+++ b/apcd_cms/src/apps/admin_extension/views.py
@@ -1,14 +1,11 @@
-from django.http import HttpResponseRedirect, HttpResponse, JsonResponse
-from django.core.paginator import Paginator, EmptyPage
+from django.http import HttpResponseRedirect, JsonResponse
 from django.views.generic.base import TemplateView
 from django.views import View
-from django.template import loader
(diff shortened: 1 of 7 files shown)


sophia-massie (approved):
Tested and LGTM 👍

happycodemonkey:
Chandra, I noticed in my testing that if you have a date set for Approved Expiration Date and clear it, and then click Submit, it will give a success message and set the date to None. When I attempted to submit and edit with no date set in another edit modal it gives an error as expected. 

Also just wanted to note that there is Python console debug logging

PR #383, opened 2024-12-04
WP-793: When log is null, handle issue gracefully instead of 500

chandra-tacc:
## Overview
1) Provide 404 error when log is null.
2) Also, currently the download log link is not available if log is null.

## Related
[WP-793](https://tacc-main.atlassian.net/browse/WP-793)

## Changes
1. Update sql to do null check.

## Testing

1. Go to http://localhost:8000/administration/view_log?log_type=html&log_id=290 to see 404.
2. Download a valid log at http://localhost:8000/administration/list-submissions/ to check for regressions

## UI
Screenshot 2024-12-04 at 9 59 11 AM


edmondsgarrett (approved):
LGTM

PR #387, opened 2024-12-04
WP-799 fix ticket creation api and handle error message

chandra-tacc:
## Overview
Registration creation - handle ticket creation error and show appropriate message
## Related
[WP-799](https://tacc-main.atlassian.net/browse/WP-799)

## Changes

1) API usage had a bug `Requestor` instead of `Requestors` 
2) wrap the call into an exception handler and give user appropriate error message.

## Testing

1.Test registration for regression: http://localhost:8000/register/request-to-submit/

### DIFF
--- a/apcd_cms/src/apps/registrations/views.py
+++ b/apcd_cms/src/apps/registrations/views.py
@@ -4,9 +4,8 @@ from apps.utils.registrations_data_formatting import _set_registration
 from apps.submitter_renewals_listing.views import get_submitter_code
 from apps.utils.apcd_groups import has_groups
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


apcd_cms/src/apps/registrations/views.py
```diff
@@ -95,15 +91,20 @@ def post(self, request):
...
-            Subject=subject,
-            Text=description,
-            Requestors=email
-        )
+        try:
+            tracker.create_ticket(
+                Queue=RT_QUEUE,
+                Subject=subject,
+                Text=description,
+                Requestor=email
```
chandra-tacc, line 100: `Requestor` instead of `Requestors`

PR #391, opened 2024-12-05
WP-796 - registration list needs to update on modal edits

chandra-tacc:
## Overview
1. The query invalidation when registration hook is successful was coded wrong. Fixed that.

## Related

[WP-796](https://tacc-main.atlassian.net/browse/WP-796)

## Changes

1. Delete un-used template
2. Sort registrations by id (desc)
3. fix query key usage. 

## Testing

2. Edit a registration, change year.
3. On success, close the modal. Check the listing - it should be updated.
4. And the listing is sorted by newest to oldest (by id)


jalowe13 (approved):
LGTM, year updates and refreshes as intended.

PR #396, opened 2024-12-06
bug/WP-797: Exception expiration year allows too large a range of dates

edmondsgarrett:
## Overview

…

## Related

- [WP-797](https://tacc-main.atlassian.net/browse/WP-123)
…

## Changes
- Used Yup's min & max functions to restrict dates on both Threshold and Exception forms to format MM/DD/YYYY
…

## Testing

1. Go to http://localhost:8000/submissions/exception/ and confirm, for both Threshold and Other exceptions, that dates not matching MM/DD/YYYY cannot be submitted (validation catches any other format and does not allow submission)
2. Submit an exception successfully and check record in Exceptions listing to confirm date is correctly formatted (regression check)

## UI

…

### DIFF
--- a/apcd_cms/src/client/src/components/Submitter/Exceptions/ExceptionFormPage.tsx
+++ b/apcd_cms/src/client/src/components/Submitter/Exceptions/ExceptionFormPage.tsx
@@ -60,7 +60,9 @@ export const ExceptionFormPage: React.FC = () => {
         .required('Business name is required'),
       fileType: Yup.string().required('File type is required'),
       fieldCode: Yup.string().required('Field code is required'),
-      expiration_date: Yup.date().required('Expiration date is required'),
+      expiration_date: Yup.date().required('Expiration date is required')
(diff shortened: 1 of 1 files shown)


sophia-massie (changes requested):
In Firefox, I'm not able to recreate this issue. It only allows four digits for the year and two valid date digits for either the month or date.

In Chrome, I am able to recreate the issue. I am able to submit invalid years still but not months or dates. It doesn't look like the max or min is catching the invalid input.

Screenshot 2024-12-06 at 4 26 47 PM

When I go to the admin listing, the exception submission is there, but with None as the Expiration Date. I'll keep debugging to see why the Yup validation isn't catching this and why the "Expiration must be MM//DD/YYYY" invalid message doesn't appear.

sophia-massie (approved):
So sorry! Totally my mistake disregard changes. Some client loading issues or cache issues on my end weren't showing validation checks and allowing bad submits. Local rebuild shows the correct behavior. Thank you!

PR #398, opened 2024-12-09
Quick Fix: APCD, List registrations for submitter is broken

chandra-tacc:
## Overview

1. Duplicate order by. Remove it.

## Testing

1. Test http://localhost:8000/administration/list-registration-requests/
2. http://localhost:8000/register/list-registration-requests/


jmartinez-tacc (approved):
LGTM!

PR #406, opened 2025-01-09
Bug: Duplicate applicable data periods in extensions form

edmondsgarrett:
## Overview
Addresses bug where a user with multiple submitters will see duplicate applicable data period options on the extensions form
…

## Related

…

## Changes
- Factors out the applicable data periods db query from collating the list of user's submitters, now keeps this list of data periods separate for each submitter
- Adds listener to business name dropdown to update list of applicable period options when business name (read: submitter) is changed
…

## Testing

1. Go to http://localhost:8000/submissions/extension-request/
  - If you only have one submitter available in the dropdown, Carrie can help add other submitters for you in our dev db
2. Check the 'Applicable Data Period' dropdown and confirm a sorted list of year-month entries are there, then select one and confirm the 'Current Expected Date' box populates with a date
3. Choose a different option under 'Business Name' and confirm that the above options change + the 'Current Expected Date' box updates with the default `mm/dd/yyyy`
4. Test with additional extensions entries and confirm they work the same + do not conflict with each other

## UI

…


sophia-massie (approved):
LGTM 👍 ! did not test locally

PR #419, opened 2025-02-03
WP-854 Updating Applicable Period Fetch Business Logic

happycodemonkey:
## Overview

Very minor update to the query that pulls applicable data periods from the submitter calendar, as described in a call with UTH where business logic was explained. See ticket referenced below for more detailed information. 

## Related

- [WP-854](https://tacc-main.atlassian.net/browse/WP-854)

## Changes

Query is updated so that it is selecting records where there is a submission id (no longer null as before), as well as cancelled = FALSE and granted_reprieve = FALSE

### DIFF
--- a/apcd_cms/src/apps/utils/apcd_database.py
+++ b/apcd_cms/src/apps/utils/apcd_database.py
@@ -1219,7 +1219,7 @@ def get_applicable_data_periods(submitter_id):
             sslmode='require',
         )
         cur = conn.cursor()
-        query = """ SELECT distinct data_period_start FROM submitter_calendar WHERE submitter_id = (%s) AND cancelled = 'FALSE' AND granted_reprieve='FALSE' AND submission_id is Null """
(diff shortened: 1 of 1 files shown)


apcd_cms/src/apps/utils/apcd_database.py
```diff
@@ -1219,7 +1219,7 @@ def get_applicable_data_periods(submitter_id):
             sslmode='require',
         )
         cur = conn.cursor()
-        query = """ SELECT distinct data_period_start FROM submitter_calendar WHERE submitter_id = (%s) AND cancelled = 'FALSE' AND granted_reprieve='FALSE' AND submission_id is Null """
+        query = """ SELECT distinct data_period_start FROM submitter_calendar WHERE submitter_id = (%s) AND cancelled = 'FALSE' AND granted_reprieve='FALSE' AND submission_id is NOT null """
```
chandra-tacc, line 1222:
```suggestion
        query = """ SELECT distinct data_period_start FROM submitter_calendar WHERE submitter_id = (%s) AND cancelled = 'FALSE' AND granted_reprieve='FALSE' AND submission_id is NOT NULL """
```

PR #420, opened 2025-02-05
WP-853: Adjust apcd backend to handle permissions error and general errors properly

chandra-tacc:
## Overview
Issues before this PR:
* Template is viewable even without permissions, and api fails with 302.
* User does not have a clear message.
* If api requests fail, then sometimes the response is not json.

## Related

[WP-853](https://tacc-main.atlassian.net/browse/WP-853)

## Changes
1. Separate template and api views in backend.
2. For template view do a redirect to '/'
3. For api view use Json Response with 403.
4. To allow api clean code, add a mixin for common auth conditions and add baseview to allow error handling

## Testing
**Follow are the page permissions:**
APCD_ADMIN: all pages
SUBMITTER ADMIN: Register menu and Submissions Menu
SUBMITTER USER: Submissions menu

**go admin pages, do the following** 
1) remove staff, super user, and any apcd groups.
2) save

**to restore:**

Use this code to enable superuser after removing access
```
python manage.py shell
```
and then
```
from django.contrib.auth import get_user_model
my_user = get_user_model().objects.get(username="<username>")
my_user.is_staff = True
my_user.is_superuser = True
my_user.save()
```

Test cases:
1) As admin all pages work
2) As submitter_user:
    Only submissions menu items work. 
    Manually try to go to 
   ```
   http://localhost:8000/administration/list-registration-requests/
   http://localhost:8000/administration/list-exceptions
   http://localhost:8000/administration/list-extensions
   http://localhost:8000/administration/list-submissions
   http://localhost:8000/register/list-registration-requests/
   http://localhost:8000/register/list-registration-requests/api
   ```
and see that you are redirected to home

3) As submitter_admin:
    See that both submissions menu and registrations work
Manually try to go to 
   ```
   http://localhost:8000/administration/list-registration-requests/
   http://localhost:8000/administration/list-exceptions
   http://localhost:8000/administration/list-extensions
    http://localhost:8000/administration/list-submissions
   ```
and see that you are redirected to home


sophia-massie (approved):
Tested and LGTM! Thanks for the work on this. We may want to remove the List Registrations from the CMS Register navigation for submitter user role if they will just be redirected, but this PR is reflecting prod so not a blocker.

chandra-tacc:
> Tested and LGTM! Thanks for the work on this. We may want to remove the List Registrations from the CMS Register navigation for submitter user role if they will just be redirected, but this PR is reflecting prod so not a blocker.

Yes, that is a good idea. That would be a CMS side change and not in this code base, can be addressed separately.

PR #431, opened 2025-02-21
bug/WP-855: Enable submit button on registration renewal load

edmondsgarrett:
## Overview

…

## Related

- [WP-855](https://tacc-main.atlassian.net/browse/WP-855)
…

## Changes
- Loads renewal data into form via Formik's `setValues` method due to Formik's `dirty` method for checking for deep differences between initial values and user input (allows submit button to be enabled on form load)
…

## Testing

1. Go to http://localhost:8000/register/list-registration-requests/ and click through a record to the 'Renew Record' action
2. Confirm that, on form load, form is populated with record's data + submit button at bottom of form is clickable
3. Submit form and confirm it submits successfully
4. Go to http://localhost:8000/administration/list-registration-requests/ and confirm there are no regressions on `Edit Record` action (submit button should not be enabled on modal load)

## UI

…


jmartinez-tacc (approved):
LGTM!

sophia-massie (approved):
LGTM! One suggestion that could be implemented site wide but could be helpful here - the button component has an isLoading property. You can add ` isLoading={registrationSubmissionPending} ` so that there is a loading spinner in side the button while the form is submitting before instead of it just being grayed out/disabled so the user knows the request is sent and working.

edmondsgarrett:
> LGTM! One suggestion that could be implemented site wide but could be helpful here - the button component has an isLoading property. You can add `isLoading={registrationSubmissionPending}` so that there is a loading spinner in side the button while the form is submitting before instead of it just being grayed out/disabled so the user knows the request is sent and working.

Thank you! I just added a commit to use this prop, good idea

PR #433, opened 2025-02-26
bug/extension-edit-modal-approved-date

sophia-massie:
## Overview

…

## Changes
- Removes string of logic checking for approved expiry date from value returned to modal for display
- Updates formatDate and formatDateUTC utils in front end to return 'None' instead of '' to fix issue with displaying 'Invalid date' when attempting to format ''
- Updates logic of displaying last updated date in View Extension Modal to match above util changes
…

## Testing

1. Go to http://localhost:8000/administration/list-extensions/
** For below steps, also check that 'Updated At:" field displays either 'None' or a date + timestamp **
2. Click through to 'Edit Record' on a record with no listed approved expiration date and confirm that only 'None' is displayed in the 'Approved Expiration Date' below the form in the modal
3. Repeat for a record with an approved expiration date and confirm date + timestamp show in the same field
4. Repeat (2) and (3) but for the 'View Record' modal

_Extra thorough step_: Check all other listing pages' 'Last Updated' column and/or View & Edit record modals to confirm either 'None' or a date+timestamp display (all listing pages but List Registration Requests and View Submitter Users display this field)

## UI
Before:
image

After:
image

…


edmondsgarrett (approved):
Nice catches here, LGTM

PR #439, opened 2025-03-03
WP-887: Fix for data periods and expected dates

chandra-tacc:
## Overview
Extension applicable data periods is using different where clause from extension expected dates resulting in periods with null expected dates.

## Related
- [WP-887](https://tacc-main.atlassian.net/browse/WP-887)

## Changes
1. Use same where in database queries for both applicable data periods and extension expected dates.
2. the api for data periods is forcing apcd admin, where as it should also allow submitter admin. Adjusted that.
3. For submitter admin, we should check whether the user has access to the submitter.
…

## Testing

1. As APCD_ADMIN, go to http://localhost:8000/submissions/extension-request/ and check all the applicable data periods
2. As SUBMITTER_ADMIN, go to http://localhost:8000/submissions/extension-request/ and check all the applicable data periods.
3. As SUBMITTER_ADMIN, open a new tab in browser and check whether you get access error when visiting this url: http://localhost:8000/common_api/data_periods/?submitter_id=0. Note: I have no access to 0, so I got `{"error": "Unauthorized for submitter id 0"}` You might have to check which submitter id you have no access to.

### DIFF
--- a/apcd_cms/src/apps/common_api/views.py
+++ b/apcd_cms/src/apps/common_api/views.py
@@ -1,8 +1,8 @@
-from django.http import HttpResponseRedirect, JsonResponse, Http404
+from django.http import JsonResponse, Http404
 from apps.utils import apcd_database
-from apps.utils.apcd_groups import has_apcd_group, is_apcd_admin
+from apps.utils.apcd_groups import is_apcd_admin
 from apps.utils.utils import title_case
-from apps.base.base import BaseAPIView, APCDAdminAccessAPIMixin, APCDGroupAccessAPIMixin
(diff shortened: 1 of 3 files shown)


apcd_cms/src/apps/common_api/views.py
```diff
@@ -67,14 +67,21 @@ def _set_cdls(cdl):
         return JsonResponse({"cdls": cdls_response})
 
 
-class DataPeriodsView(APCDAdminAccessAPIMixin, BaseAPIView):
```
chandra-tacc, line 70: This is a bug

sophia-massie (commented):
Unrelated but a quick fix - In ExtensionsForm component, the request and justification description says "exception" where it should say extension.

sophia-massie (approved):

happycodemonkey (approved):
LGTM

PR #461, opened 2025-04-03
Removed the json log link in the file submissions modal

happycodemonkey:
## Overview

UTH request, the JSON logs are for internal use only. Removed from both Admin and User modals, since I assume internal use means it should not be displayed in the portal in general. 

## Related

- [WP-914](https://tacc-main.atlassian.net/browse/WP-914)

## Changes

Removed JSON link from the modal template. 

## Testing

1. Open http://localhost:8000/submissions/list-submissions/ and verify the JSON link is no longer displayed 
2. Do the same on http://localhost:8000/administration/list-submissions/

### DIFF
--- a/apcd_cms/src/client/src/components/Submissions/ViewFileSubmissions/ViewSubmissionsModal.tsx
+++ b/apcd_cms/src/client/src/components/Submissions/ViewFileSubmissions/ViewSubmissionsModal.tsx
@@ -70,20 +70,6 @@ export const ViewSubmissionLogsModal: React.FC<
                           </Link>
                         )}
                       </dd>
-                      <dt className="c-data-list__key">JSON Log</dt>
-                      <dd className="c-data-list__value">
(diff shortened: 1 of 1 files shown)


sophia-massie (approved):
Tested and LGTM

edmondsgarrett (changes requested):
We do still need this link, but just for the admin listing - need to hide for the user listing.

Looks like we could use the `isAdminUser` for this, possibly?

sophia-massie:
> We do still need this link, but just for the admin listing - need to hide for the user listing.
> 
> Looks like we could use the `isAdminUser` for this, possibly?

I think Tracy said [in Slack](https://tacc-team.slack.com/archives/C07JXANHJUS/p1743710022106579) to remove from both admin and submitter pages. Here is the thread from that discussion

edmondsgarrett (approved):
LGTM!

PR #463, opened 2025-04-07
task/Points v1.0.19 (prod cms tag) to TACC CMS image v4.26.1

sophia-massie:
## Overview

Updates prod CMS to keep everything the same as v1.0.19 (prod CMS) with no react client but points base image to CoreCMS v4.26.1

## Changes

…

## Testing

1. Make build / make start
2. Ensure it looks like https://txapcd.org/ 

## UI

…

### DIFF
--- a/apcd_cms/Dockerfile
+++ b/apcd_cms/Dockerfile
@@ -1,7 +1,8 @@
-FROM taccwma/core-cms:v4.12.0
+FROM taccwma/core-cms:v4.26.1
 
 WORKDIR /code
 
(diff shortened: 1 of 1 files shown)


rstijerina (approved):

edmondsgarrett (changes requested):
On registration form, max-width rule on many of the inputs is no longer on them, so they extend the full width of the form:
image

Did either of y'all see this?

edmondsgarrett (approved):
Seems like above issue is likely a local dev issue only, LGTM

PR #475, opened 2025-04-25
Hotfix: Null created_at value

happycodemonkey:
Fixed issue where admin extensions page created null datetime value that was wrong type.


sophia-massie (approved):
🎉 Thank you so much for this fix!

PR #483, opened 2025-05-09
View/Download Logs Bug

happycodemonkey:
## Overview

Fixing small issue with case statement, see code change.


sophia-massie (approved):
LGTM. Should we also remove the json content_type  in the get_log function in admin_submissions view.py [here](https://github.com/TACC/Core-CMS-Custom/blob/77557b54689b60739474ce712aee2dc08a97dba5/apcd_cms/src/apps/admin_submissions/views.py#L133?

happycodemonkey:
> LGTM. Should we also remove the json content_type in the get_log function in admin_submissions view.py [here](https://github.com/TACC/Core-CMS-Custom/blob/77557b54689b60739474ce712aee2dc08a97dba5/apcd_cms/src/apps/admin_submissions/views.py#L133)?

Removed the json content type!

sophia-massie (approved):

PR #491, opened 2025-05-20
Feature/wp 926 admin registration list facets

davidmtacc:
## Overview
Added COLUMN 'Created' showing date and time of creation

## Related

…

## Changes

Change Location in table to Created and used what was available from query to replace with date/time requested

## Testing

1. Go to /administration/list-registration-requests/
2.  Verify Column after Type shows Created with Date/Time

## UI

Looks and feels the same only for substitution of Location column to Created column

## Notes
Since the table on this page draws data from the same table under Register/list-registration-requests I had planned to send them both up together but for some reason I cannot get list-registration-requests to load on my local.  
Also, second part of the task was to add user who created this but there was confusion as to which user.
…


sophia-massie:
Moved WIP to new repo [/TACC/APCD-CMS](https://github.com/TACC/APCD-CMS)

